PR #5, opened 2022-04-08
Added "Zero PII" section

makwarth:
Also added note about reserved attributes (name)


README.md
```diff
@@ -50,11 +50,28 @@ Supply these to the `init` call (2nd argument)
...
+
+bucket.user(await sha256("john_doe"));
+```
+
 ### Custom attributes
 
 You can pass attributes as a object literal to the `user`, `company` and `track` methods (2nd argument).
 Attributes cannot be nested (multiple levels) and must be either strings, integers or booleans.
 
+Reserved attributes:
```
laander, line 71: Maybe not "reserved" since it sounds like something you can't use. "Built-in" ?

PR #8, opened 2022-12-30
add support for sending user feedback

roncohen:


src/main.ts
```diff
@@ -123,6 +124,42 @@ export default function main() {
     return res;
   }
 
+  type FeedbackOptions = {
+    featureId: string;
+    sentiment: "like" | "dislike";
+    userId?: string;
+    companyId?: string;
+    comment?: string;
+  };
```
laander, line 133: This looks identical to the `Feedback` type in `src/types.ts` - perhaps just get rid of this?

src/main.ts
```diff
@@ -123,6 +124,42 @@ export default function main() {
...
+    featureId,
+    sentiment,
+    userId,
+    companyId,
+    comment,
+  }: FeedbackOptions) {
+    checkKey();
+    if (!featureId) err("No featureId provided");
+
+    const payloadUserId = userId || getSessionUser();
```
laander, line 144: Should check for `if (persistUser)` before retrieving `getSessionUser()` like in the other methods like `track` above

src/main.ts
```diff
@@ -123,6 +124,42 @@ export default function main() {
...
+    if (!payloadUserId) {
+      err("No userId provided and persistUser is disabled");
+      return;
+    }
+
```
laander, line 149: Given that `sentiment` is a required parameter, we should probably add runtime check for it too, like
```
if (!sentiment) err("No sentiment provided");
```
Should sentiment be required tho? Isn't it possible to add a written feedback (note) without proving a like or dislike?

src/main.ts
roncohen: this one is slightly better. Might do that

Screenshot 2023-01-09 at 14 11 22

src/main.ts
roncohen: no way to send feedback without a sentiment in this version. We can always make it more lax in future iterations.

src/main.ts
roncohen: I like this version because `getSessionUser` will always return a string or throw, so there's no reason to check persistUser separately. I don't want to do that refactor now though, so I've just aligned with how we do it otherwise. I did get rid of the exclamation mark operator [here](https://github.com/bucketco/bucket-tracking-sdk/blob/main/src/main.ts#L95 and [here](https://github.com/bucketco/bucket-tracking-sdk/blob/main/src/main.ts#L117) by setting the return value for `err` to `never`: https://github.com/bucketco/bucket-tracking-sdk/pull/8/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR48

Munter (approved):

PR #21, opened 2023-06-27
add context.active

roncohen:


Munter (approved):
The changes look fine

I'm a bit in doubt about whether we should have the client send explicit context to the API, so a non-definition of `context` or `context.active` would result in `context: { active: true }` in the request. Doing so would make it more clear what's happening, but on the other hand we can't control the default in a central location in ingest or the message processor

roncohen:
that is definitely an interesting point. I think this will probably be OK for now. If it turns out to be confusing, we can add it so it's explicit later on 👍

roncohen:
updated the README to include context.active=false

PR #24, opened 2023-07-31
Feedback modal using Preact and vanilla CSS

Swiftwork:
Based on details from here
https://www.notion.so/bucketdotco/Feedback-widget-RFC-e700c1c38c684605b519131b120747f3
and here
https://www.notion.so/bucketdotco/Embeddable-feedback-widget-37187ab4ac3a4abcab173709008d5be2.
Also implements floating UI library https://floating-ui.com/docs/getting-started

image
image
image


src/feedback/index.css
```diff
@@ -0,0 +1,5 @@
+@import url(./Spinner.css);
+@import url(./Button.css);
+@import url(./FeedbackDialog.css);
+@import url(./FeedbackForm.css);
+@import url(./StarRating.css);
```
Munter, line 5: Flattening out the dependency graph into a manifest like this is eventually going to hurt us. I think we should import component CSS from the component the CSS affects to avoid mistakes with forgetting to import CSS here, or removing it again when a component is no longer imported

src/main.ts
```diff
@@ -181,6 +182,29 @@ export default function main() {
     return res;
   }
 
+  function collectFeedback(options: FeedbackDialogOptions) {
```
Munter, line 185: The SDK is used by both node and browser. But this method doesn't make sense anywhere else than a browser.

Potential solutions:
- Throw an explicit error in node (calling the function will already throw in node, but not with a good error message)
- Separate builds with separate API's for node and browser

I think I prefer throwing for now, to avoid getting into all sorts of dual configuration, deployment and documentation work

src/feedback/index.css
Swiftwork: I'm all for that, just ran into an issue where the styles get duplicated if you for example use the button component multiple times. Happy to do it either way but we need to dedupe the styles somehow.

Munter: Ah, a conflict with the need for inlining the styles I presume. Maybe we can save this for later then

README.md
```diff
@@ -37,14 +37,22 @@ bucket.company("acme_inc", { name: "Acme Inc", plan: "pro" }, "john_doe");
...
 bucket.feedback({
   featureId: "my_feature_id",
   userId: "john_doe",
   companyId: "acme_inc", // String (optional)
   score: 5, // Number: 1-5 (optional)
   comment: "Absolutely stellar work!", // String (optional)
 });
+
+// collect qualitative feedback using built-in dialog
+bucket.collectFeedback({
```
roncohen, line 50: quick drive-by: let's call this `openFeedbackForm` or `openFeedbackDialog` to make it super explicit. `collectFeedback` could mean _send_ the feedback.

roncohen:
we'd like the first version to support translations in some way. A very simple way to achieve that is to make is possible to pass the texts in from outside, like "How did you like FeatureA? (`feedbackQuestion`?), "Leave a comment" (`commentTitle`?), "Thank you for sending feedback!" (`thankYouText`?) and so on.

rrpff:
Makes sense @roncohen - have a separate ticket to address that, will merge this initial (undocumented) PR without it.

Munter (approved):
The new additions look good. I think we are ready to go

Munter:
I fixed the linting config problems, now there are a few actual linting errors left

src/main.ts
```diff
@@ -324,6 +324,10 @@ export default function main() {
   }
 
   async function openFeedbackForm(options: FeedbackDialogOptions) {
+    if (typeof window === "undefined") {
+      err("openFeedbackForm can only be called in the browser");
```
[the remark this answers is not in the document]
rrpff: The `err` function does throw when outside of debug mode 👍

PR #25, opened 2023-08-10
Enable ably integration

pavkam:
This PR introduces experimental support for Ably as our provider of WebSockets. We also present the `feedback prompting` concept that can be configured to be either automatic or manually triggered.

Things of note:
- The `host` setting of Ably is not set yet to `tracking.bucket.com` - that will need to be configured first,
- The UX for feedback prompting is not merged yet so the current default method is a stub.

Both will need to be addressed at a later PR.

### DIFF
--- a/src/ably.ts
+++ b/src/ably.ts
@@ -1,7 +1,6 @@
 import Ably from "ably/promises";
-import { ABLY_CHANNEL } from "./config";
 
-export async function openAblyConnection(authUrl: string, userId: string, callback: (req: any) => void, debug?: boolean) {
+export async function openAblyConnection(authUrl: string, userId: string, channel: string, callback: (req: any) => void, debug?: boolean) {
   const client = new Ably.Realtime.Promise({
     authUrl: authUrl,
     log: { level: debug ? 4 : 1 },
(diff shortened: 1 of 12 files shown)


src/main.ts
```diff
@@ -181,8 +205,61 @@ export default function main() {
...
+    }, debug)
+
+    log(`feedback prompting connection established`);
+    return res;
+  }
+
+  function showFeedbackPrompt(prompt: FeedbackPrompt) {
+    // nothing to do here yet
+    log(`showing feedback prompt`, prompt);
+  }
```
roncohen, line 253: lets just remove this for now? no need to have that in here. It's totally fine to ship prompting without the embeddable widget.

roncohen (approved):
let's wait a bit with the documentation. Otherwise LGTM!

src/ably.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function openAblyConnection(authUrl: string, userId: string, channel: string, callback: (req: any) => void, debug?: boolean) {
+  const client = new Ably.Realtime.Promise({
+    authUrl: authUrl,
+    log: { level: debug ? 4 : 1 },
+    authParams: {
+      userId,
+    }
+  });
+  const c = client.channels.get(channel);
```
matus-vacula, line 11: Here you will have to use the rewind parameter to retrieve the last persisted message. https://ably.com/docs/channels/options/rewind

PR #26, opened 2023-08-21
Add linting and upgrade dependencies

pavkam:
- Added ESLit and Prettier,
- Cleaned up code a bit (types),
- Upgraded most dependencies and tested all work fine,
- Improved the GH action.


package.json
```diff
@@ -8,7 +8,11 @@
     "build": "tsc && webpack",
     "test": "vitest",
     "test:ci": "vitest --reporter=junit --outputFile=junit.xml",
-    "coverage": "vitest run --coverage"
+    "coverage": "vitest run --coverage",
+    "lint": "eslint .",
+    "lint:ci": "eslint --output-file eslint-report.json --format json .; exit 0",
```
Munter, line 13: Won't `exit 0` just make this task never fail?
I would think we'd want the pipeline to fail if linting fails

Munter (approved):
Looks great!

I like the small change to the debug loggin to ensure the message comes before the rest of the payload

package.json
pavkam: Good question! I just blindly copy-pasted it from our ingest pipeline

pavkam: I think that might have been a work-around in the past to not fail the reporting step. But that is taken care of now with custom conditions. Will update.

PR #32, opened 2023-08-25
Add E2E tests around built SDK

rrpff:


package.json
```diff
@@ -10,8 +10,9 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc && webpack",
-    "test": "vitest",
-    "test:ci": "vitest --reporter=junit --outputFile=junit.xml",
+    "test": "vitest -c vite.config.js",
```
pavkam, line 13: add this `vitest -c vite.config.js` to `preversion` as well

test/e2e/acceptance.node.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { randomUUID } from "crypto";
+import { beforeEach, test } from "vitest";
+import nock from "nock";
+import bucket from "../../";
+
+const KEY = randomUUID();
+
+beforeEach(() => nock.cleanAll());
+
+test("Acceptance", async () => {
```
pavkam, line 11: you want to add a `nock.isDone()` to check that all listeners have bit hit

test/e2e/acceptance.browser.spec.ts
```diff
@@ -0,0 +1,96 @@
+import { test, expect } from "@playwright/test";
```
pavkam, line 1: Interesting, what does this testing bring us on top of what we have now? We already use `nock` in our unit tests so we are testing that each call actually goes out the library.

pavkam:
As I expressed in a comment above. What is the benefit of using playwright here over what we have already? As far as I can see you are testing the same code-paths with the same outcomes.

rrpff:
@pavkam yeah, the test's are extremely similar - I actually copied the mocks, even. However these tests use the `dist/` version of the are run within the environment they'll be used in.

Here I'm just looking for baseline proof that at least:
- the node build can be imported into node without errors
- the browser build can be imported into common browsers (chrome,firefox,safari) without errors
- all the basic functions work in both environments

IMO vitest is subtly different enough from both `node` and from browsers to make this worth the duplication / extra build time. For example, #24 currently breaks the node build without failing any tests.

pavkam:
> @pavkam yeah, the test's are extremely similar - I actually copied the mocks, even. However these tests use the `dist/` version of the are run within the environment they'll be used in.
> 
> 
> 
> Here I'm just looking for baseline proof that at least:
> 
> - the node build can be imported into node without errors
> 
> - the browser build can be imported into common browsers (chrome,firefox,safari) without errors
> 
> - all the basic functions work in both environments
> 
> 
> 
> IMO vitest is subtly different enough from both `node` and from browsers to make this worth the duplication / extra build time. For example, #24 currently breaks the node build without failing any tests. 

That makes sense! It won't cover the feedback prompting though - that's a tough cookie.

Munter:
@rrpff I pushed some updates to make the linting setup run correctly, which also removes the inclusion of the webpack config file in the dist folder. Locally, this seems to fix all the problems I can find on this branch.

Great to have the browser tests in there!

rrpff:
Thanks @Munter ! Not sure what's up with my local linting setup, will investigate

PR #34, opened 2023-08-28
Create dependabot.yml

pavkam:


matus-vacula:
Do we actually want this if we decided that we'll rather do it once every 3 months?

pavkam:
> Do we actually want this if we decided that we'll rather do it once every 3 months?

Good point. Maybe not, if we will keep it up to date ourselves.

PR #39, opened 2023-08-31
Replace Ably SDK with pure SSE

pavkam:
BUC-1357

Much leaner code. SDK is now 98Kb.


src/sse.ts
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    this.err("server did not release a token request", res);
+  }
+
+  private async refreshToken() {
+    const tokenRequest = await this.refreshTokenRequest();
+
+    const res = await fetch(
+      `${ABLY_REST_HOST}/keys/${encodeURIComponent(
```
matus-vacula, line 77: You can make this overridable with options and already set the default host to the bucket CNAME.

matus-vacula (approved):
Looks great 👍

src/sse.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    this.err("server did not release a token", res);
+  }
+
+  private async onError(e: Event) {
+    if (e instanceof MessageEvent) {
+      const errorPayload = JSON.parse(e.data);
+      const errorCode = Number(errorPayload?.code);
+
+      if (errorCode >= 40140 && errorCode < 40150) {
```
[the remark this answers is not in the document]
pavkam: Yeah, I took that from their documentation and their example code. I've searched the docs now for error codes and added costs. Thanks.

src/sse.ts
pavkam: Initially, I made them configurable, but then I removed the options since Ably does not actually have any other endpoints per environment so we will never really override these. So making them overridable is not really useful. Also, this client is Ably-specific so we would need to create a new client for any other back-end system we will have in the future.

We don't yet know the CNAME, still waiting for clarifications, so will only release the new SDK with those new CNAMEs plugged in and that will be it :)

PR #41, opened 2023-09-04
Misc changes required for internal usage

rrpff:
Closes BUC-1375
Closes BUC-1344
Closes BUC-1342
Closes BUC-1337


rrpff:
nice! merging :)

src/default-feedback-prompt-handler.ts
```diff
@@ -0,0 +1,13 @@
+import { FeedbackPrompt, FeedbackPromptHandler } from "./types";
```
pavkam, line 1: would love a test of `defaultFeedbackPromptHandler` and another test that checks that `defaultFeedbackPromptHandler` is called by feedback mechanism if no user callback is assigned.

PR #43, opened 2023-09-07
Support translating strings within the widget

rrpff:
- Allows overriding every string in the widget
- _except_ for the "Powered by Bucket" text.
- I removed the `dimmed` of the "(optional)" text above the comment. Looks a bit less nice, but felt awkward having to provide two translation keys here

image


Munter (approved):
Les Français remercient pour la reconnaissance :)

rrpff:
mais bien sûr !

PR #44, opened 2023-09-07
Set a long expiry on prompt cookies to prevent repeat showings

rrpff:


Munter:
I don't think this is a good idea. We don't want to litter a customers domain with cookies for years on end. We might actually risk flooding them so much that they reach the cookie limit.

I think the original code is correct; the cookie should expire when the prompt expires, maybe with an adjustment with an extra day for the people who change their clock rather than their time zone when they travel.

If we do pop up the prompt beyond that, I think we have a bug somewhere else that we should fix

rrpff:
In this case it's a single cookie per user id, so think it's unlikely we'd reach a cookie limit. You're right though that this issue should also be fixed by the `showBefore` bug, looking at that now.

PR #50, opened 2023-09-21
Block all event propagation across the widget container boundary

Munter:


src/feedback/constants.ts
```diff
@@ -1 +1,65 @@
 export const feedbackContainerId = "bucket-feedback-dialog-container";
+
+// see https://developer.mozilla.org/en-US/docs/Web/API/Element#events
+export const nonPropagatedEvents = [
```
Swiftwork, line 4: The name of this const is a little confusing as I assume it is a list of events that do propagate but that we don't want to. I would change the name to `propagatingEvents` as that is more descriptive of the list even if it is used to stop propagation. Maybe also move it to the shared project.

Swiftwork (changes requested):

src/feedback/constants.ts
Munter: Name changed.

This is only being used in the feedback ui though, so I think it's better to keep it here for now

PR #51, opened 2023-09-26
Send "featureId" parameter when submitting prompt events

rrpff:


Munter (approved):

src/main.ts
```diff
@@ -325,19 +340,21 @@ export default function main() {
     feedbackPromptHandler?.(message, handlers);
   }
 
-  async function feedbackPromptEvent(
-    promptId: string,
-    event: "received" | "shown" | "dismissed",
-    userId: User["userId"],
-  ) {
+  async function feedbackPromptEvent(options: {
```
pavkam, line 343: nit: these are not technically `options` but `args`

pavkam (approved):
LGTM

src/main.ts
rrpff: fair point 👍

PR #53, opened 2023-09-26
Rework puclic API - better ui config in bucket.init()

Munter:


src/types.ts
```diff
@@ -39,12 +58,60 @@ export type TrackedEvent = {
...
+  featureId: string;
+  userId: string;
+  companyId?: string;
+
+  /**
+   * Allows you to handle a copy of the already submitted
+   * feedback.
+   *
+   * This can be used for side effects, such as storing a
+   * copy of the feedback in your own applicaiton or CRM.
```
pavkam, line 72: `applicaiton` -> `application`

src/types.ts
```diff
@@ -39,12 +58,60 @@ export type TrackedEvent = {
...
+}
+
 export type Feedback = {
+  /**
+   * Bucket feature ID
+   */
   featureId: string;
+
+  /**
+   * User id from your own appliction
```
pavkam, line 87: `appliction` -> `application`

pavkam (approved):

src/types.ts
Munter: I'll fix those before cutting a release

src/types.ts
Munter: I'll fix those before cutting a release

PR #58, opened 2023-10-02
Disable prompting on mobile

rrpff:
Super naive implementation. I felt a viewport check was suitable enough, given the reason we're disabling on mobile is that it needs some work on small screens.


Munter (approved):
Looks good enough for our current needs 👍

PR #63, opened 2023-10-03
Brand prompting as "Live Feedback"

rrpff:


laander (commented):
Let's try always coupling `live` to `feedback`, as it can be hard to understand on it's own. So `liveFeedback`

roncohen (approved):

PR #67, opened 2023-10-18
Slim Feedback Widget

rrpff:


src/feedback/FeedbackForm.tsx
```diff
@@ -56,53 +68,125 @@ export const FeedbackForm: FunctionComponent<FeedbackFormProps> = ({
...
+        submittedRef.current!.style.position = "relative";
+        submittedRef.current!.style.opacity = "1";
+        submittedRef.current!.style.pointerEvents = "all";
+        setShowForm(false);
+      }, ANIMATION_SPEED + 10);
+    } else {
+      const isExpanded = openWithCommentVisible || hasRating;
+
+      containerRef.current.style.maxHeight = isExpanded
+        ? "400px" // TODO: reconsider?
```
laander, line 101: Unfortunately, the way `maxHeight` and transitions work means that the size just gets clamped and the transition won't end at the actual height but instead the maxHeight. The result being that the transition stops abruptly. 
Perhaps there's another way we can achieve this resizing while still take dynamic question text height into account?

laander (commented):
Took the liberty to make a few tweaks directly in code FYI 🤠

src/feedback/FeedbackForm.tsx
rrpff: Definitely, this has been on my list to sort out. I'm currently calculating the exact size of the initial screen (before it expands to show the comment) and of the success screen, so was planning on using the same trick here for the full expanded screen. i.e. setting max-height to (score div height + comment div height + gap height) should work 🤞

src/feedback/FeedbackDialog.css
```diff
@@ -36,8 +36,8 @@
 
 .dialog {
   position: fixed;
-  width: max-content;
-  padding: 16px;
+  width: 274px;
+  padding: 16px 22px 10px;
```
Swiftwork, line 40: Very specific sizes and paddings. Don't you think it would be wise to match out `4` based system `4, 8, 12, 16, 20, 24`?

src/feedback/FeedbackForm.css
```diff
@@ -42,17 +68,44 @@
     Open Sans,
     sans-serif
   );
-  resize: vertical;
+
+  resize: none;
```
Swiftwork, line 72: This could be frustrating if someone wants to write a lot.

src/feedback/StarRating.tsx
```diff
@@ -1,97 +1,180 @@
...
-    icon: <VerySatisfied />,
-    getLabel: (t: FeedbackTranslations) => t.ScoreVerySatisfiedLabel,
+    color: "var(--bucket-feedback-dialog-rating-1-color, #dd6b20)",
+    bg: "var(--bucket-feedback-dialog-rating-1-background-color, #fbd38d)",
+    icon: <VeryDissatisfied />,
+    getLabel: (t: FeedbackTranslations) => t.ScoreVeryDissatisfiedLabel,
+    value: 1,
   },
-];
+] as const;
```
Swiftwork, line 48: https://linear.app/bucket/issue/BUC-1336/investigate-slim-version-of-modal-which-asks-for-score-then-comment#comment-addd4395

Swiftwork (approved):
Minor questions but nothing that majorly stood out to me except for the reversed star ratings.

src/feedback/FeedbackDialog.css
rrpff: Good point, tweaked the width to `280` - I left the padding as is though, the `10` is because we apply `6px` elsewhere

src/feedback/FeedbackForm.css
rrpff: Unfortunately necessary for now, since the UX of resizing a textarea that's a fixed height from the bottom right felt super weird. We could invest in an auto-resizing textarea in future maybe?

src/feedback/StarRating.tsx
rrpff: Changed 👍

PR #71, opened 2023-10-30
Live Feedback -> Live Satisfaction

rrpff:


pavkam (commented):
LGTM

pavkam (approved):

Munter (approved):
Looks good.

Since we're going with the deprecation route, I assume we can get away with just a minor version here.

I notice that the deprecated, but still functional paths, are not being tested any more. Should we keep some of those tests around to ensure they work?

rrpff:
@Munter - good point, I had a play with using `describe.each` to test both methods, but it got pretty clunky. I'll think about this in a future PR, want to get this merged for now.

PR #75, opened 2023-11-02
Improve retries and add more guarding code

pavkam:
This PR re-arranges some code to move "get a new token" retry from inside the "event listener's" loop and join it to the main retry loop.

This is to avoid any hypothetical situation when Ably misbehaves about our tokens' validity and we constantly try to obtain a new token and connect to Ably, thus entering an unwanted loop.

To improve things, I have also made sure that when we connect successfully, we reset the retry count to its default value to not exhaust retries on normal reconnects due to token expiration.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -287,6 +287,19 @@ export default function main() {
 
     log(`feedback prompting enabled`);
 
+    if (sseChannel) {
+      if (feedbackPromptingUserId !== userId) {
+        err(
+          "Feedback prompting already initialized for a different user. Are you calling initLiveSatisfaction() or user() in a loop?",
+        );
+      } else {
+        warn(
+          "Feedback prompting already initialized for this user. Are you calling initLiveSatisfaction() or user() in a loop?",
+        );
+        return res;
+      }
+    }
+
     feedbackPromptingUserId = userId;
     sseChannel = openAblySSEChannel(
       `${getUrl()}/feedback/prompting-auth`,
--- a/src/main.ts
+++ b/src/main.ts
@@ -47,6 +47,7 @@ export default function main() {
   let trackingHost: string = TRACKING_HOST;
   let sessionUserId: string | undefined = undefined;
   let persistUser: boolean = !isForNode;
+  let sseChannelActive: boolean = false;
   let sseChannel: AblySSEChannel | undefined = undefined;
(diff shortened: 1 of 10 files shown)


src/main.ts
```diff
@@ -287,6 +287,19 @@ export default function main() {
 
     log(`feedback prompting enabled`);
 
+    if (sseChannel) {
```
roncohen, line 290: there's already a check here: https://github.com/bucketco/bucket-tracking-sdk/pull/75/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR267-R269

pavkam: True, but:
```
const res = await request(`${getUrl()}/feedback/prompting-init`, {
      userId,
    });
```
will split the function into 2+ promises which leads to a race condition. Since calling `user("alex"); user("john");` not being awaited, will both bypass that check.

pavkam: made the changes as discussed

src/main.ts
```diff
@@ -47,6 +47,7 @@ export default function main() {
   let trackingHost: string = TRACKING_HOST;
   let sessionUserId: string | undefined = undefined;
   let persistUser: boolean = !isForNode;
+  let sseChannelActive: boolean = false;
```
roncohen, line 50: maybe?
```suggestion
  let initLiveSatisfactionInitialized: boolean = false;
```

rrpff (approved):
Changes look sound to me 👍

src/main.ts
pavkam: Done

src/main.ts
```diff
@@ -275,28 +274,38 @@ export default function main() {
...
-    const res = await request(`${getUrl()}/feedback/prompting-init`, {
-      userId,
-    });
-    log(`feedback prompting status sent`, res);
-    const body: { success: boolean; channel?: string } = await res.json();
-    if (!body.success || !body.channel) {
-      log(`feedback prompting not enabled`);
-      return res;
-    }
+    // while initializeing, consider the channel active
```
matus-vacula, line 277:
```suggestion
    // while initializing, consider the channel active
```

matus-vacula (approved):
Looks good!

PR #76, opened 2023-11-02
Visual widget tweaks (width, line-height)

laander:
Two minor tweaks:
- Make the widget a tad slimmer in width by default since there's a lot of blank whitespace on the right, which makes it feel very left-heavy. Cons: it makes for less space in the question before it wraps + the feedback textarea becomes a bit smaller. I think both are acceptable though as it still allows for decent length of the question before wrapping to 2+ lines.
- Increase line-height somewhat for the title and textarea to give it slightly more breathing room

https://github.com/bucketco/bucket-tracking-sdk/assets/222419/1078f2eb-73e2-4eae-aecf-df0b5ccc858d

### DIFF
--- a/src/feedback/FeedbackDialog.css
+++ b/src/feedback/FeedbackDialog.css
@@ -36,7 +36,7 @@
 
 .dialog {
   position: fixed;
-  width: 280px;
+  width: 210px;
   padding: 16px 22px 10px;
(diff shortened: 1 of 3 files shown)


rrpff (approved):
LGTM 👍

PR #78, opened 2023-11-14
Feedback Source

rrpff:


Swiftwork (approved):
Out of curiosity and education, can you write a brief comment description for when we can expect each source?

```
"api" | "manual" | "prompt" | "sdk" | "widget"
```

src/types.ts
```diff
@@ -138,6 +138,11 @@ export type Feedback = {
    * feedback.
    */
   promptId?: FeedbackPrompt["promptId"];
+
+  /**
+   * Source, depending on how the user was asked
+   */
+  source?: "api" | "manual" | "prompt" | "sdk" | "widget";
```
pavkam, line 145: Should we remove `api` and `manual` from this enum? The SDK does not use them and probably better not to be advertised.

rrpff: Yep, good shout

rrpff:
@Swiftwork good shout - added some extra info to the comment. I don't expect this property to be used by customers, but didn't want to introduce extra misdirection to avoid it being visible.

PR #79, opened 2023-11-16
feat: allow custom SSE host

roncohen:
`init` options now take a separate `sseHost` to let users select a different host for the live connection. This is in addition the existing `host` option. Both of these are useful for proxying.


pavkam (approved):
LGTM

pavkam (commented):
The only thing to note that `sseHost` is used for both `SSE` and `REST` calls, but I don't think people can selectively disable REST if proxying as they both do GET requests.

roncohen:
Thanks!

PR #85, opened 2023-12-06
Add JSDoc documentation to SDK methods

rrpff:


src/main.ts
```diff
@@ -103,6 +103,14 @@ export default function main() {
     return userId!;
   }
 
+  /**
+   * Initialize the Bucket SDK.
+   *
+   * Must be called before calling other SDK methods.
+   *
+   * @param key Your Bucket tracking key
+   * @param options
```
pavkam, line 112: should we describe each option here?

src/main.ts
```diff
@@ -151,6 +159,13 @@ export default function main() {
     log(`initialized with key "${trackingKey}" and options`, options);
   }
 
+  /**
+   * Identify the current user in Bucket, so they are tracked against each `track` call and receive Live Satisfaction events.
+   *
+   * @param userId The ID you use to identify this user
+   * @param attributes Any attributes you want to attach to the user in Bucket
+   * @param context
```
pavkam, line 167: same here for Context and attributes, can we add type descriptions?

src/types.ts
```diff
@@ -69,8 +69,19 @@ export type TrackedEvent = {
 
 export interface RequestFeedbackOptions
   extends Omit<OpenFeedbackFormOptions, "key" | "onSubmit"> {
+  /**
+   * Bucket feature ID
+   */
   featureId: string;
+
+  /**
+   * User id from your own application.
```
pavkam, line 78: Upper case the ID

src/types.ts
```diff
@@ -69,8 +69,19 @@ export type TrackedEvent = {
...
+   */
   userId: string;
+
+  /**
+   * Company id from your own application.
```
pavkam, line 83: capitalize ID

src/main.ts
```diff
@@ -442,6 +488,13 @@ export default function main() {
     return res;
   }
 
+  /**
+   * Display the Bucket feedback form UI programmatically.
+   *
+   * This can be used to collect feedback from users in Bucket in cases where Live Satisfaction isn't appropriate.
+   *
+   * @param options
```
pavkam, line 496: option type descritption

src/types.ts
```diff
@@ -69,8 +69,19 @@ export type TrackedEvent = {
 
 export interface RequestFeedbackOptions
   extends Omit<OpenFeedbackFormOptions, "key" | "onSubmit"> {
+  /**
```
pavkam, line 72: should we annotate all types in this file? And can we reference them in the `main` module to describe those `options` and etc?

rrpff:
@pavkam I didn't describe each option when there are object args, as there are already JSDoc descriptions on the properties themselves. I originally started down that road, but felt the likelihood of the duplication drifting was too high.

Personally when using IDE hints, I only read the function description for general usage, and then each option for more specificity, so went with that approach.

i.e. function:
image

property:
image

Thoughts?

pavkam:
> @pavkam I didn't describe each option when there are object args, as there are already JSDoc descriptions on the properties themselves. I originally started down that road, but felt the likelihood of the duplication drifting was too high.
> 
> Personally when using IDE hints, I only read the function description for general usage, and then each option for more specificity, so went with that approach.
> 
> i.e. function: image
> 
> property: image
> 
> Thoughts?

I know some people prefer pure JSDoc to TS. I am by no means an expert in how these things work to be honest. I'll defer to your judgement.

pavkam (approved):

rrpff:
Not an expert either, noted prisma/react/chakra/react-router just have a short description then link to docs. Will merge this for now as it's an improvement

PR #88, opened 2023-12-12
Cache Ably auth-cookie

pavkam:
* Cache the cookie locally (for its validity time),
* Also, added cookie settings to avoid Firefox complaints: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -27,6 +27,7 @@ import type {
   TrackedEvent,
   User,
 } from "./types";
+import { getAuthToken } from "./prompt-storage";
 
 async function request(url: string, body: any) {
   return fetch(url, {
@@ -324,26 +325,33 @@ export default function main() {
 
     userId = resolveUser(userId);
 
+    const existingAuth = getAuthToken(userId);
+    let channel = existingAuth?.channel;
+
     // while initializing, consider the channel active
     liveSatisfactionActive = true;
     try {
-      const res = await request(`${getUrl()}/feedback/prompting-init`, {
-        userId,
-      });
+      if (!channel) {
+        const res = await request(`${getUrl()}/feedback/prompting-init`, {
+          userId,
+        });
 
-      log(`feedback prompting status sent`, res);
-      const body: { success: boolean; channel?: string } = await res.json();
-      if (!body.success || !body.channel) {
-        log(`feedback prompting not enabled`);
-        return res;
+        log(`feedback prompting status sent`, res);
+        const body: { success: boolean; channel?: string } = await res.json();
+        if (!body.success || !body.channel) {
+          log(`feedback prompting not enabled`);
+          return res;
+        }
+
+        channel = body.channel;
       }
 
-      log(`feedback prompting enabled`);
+      log(`feedback prompting enabled`, channel);
 
       sseChannel = openAblySSEChannel(
         `${getUrl()}/feedback/prompting-auth`,
         userId,
-        body.channel,
+        channel,
         (message) => handleFeedbackPromptRequest(userId!, message),
         { debug, sseHost },
       );
@@ -351,11 +359,11 @@ export default function main() {
       feedbackPromptingUserId = userId;
 
       log(`feedback prompting connection established`);
-      return res;
     } finally {
       // check that SSE channel has actually been opened, otherwise reset the value
       liveSatisfactionActive = !!sseChannel;
     }
+    return channel;
   }
 
   function handleFeedbackPromptRequest(userId: User["userId"], message: any) {
--- a/src/main.ts
+++ b/src/main.ts
@@ -7,6 +7,7 @@ import type { FeedbackPosition, FeedbackTranslations } from "./feedback/types";
 import { SSE_REALTIME_HOST, TRACKING_HOST } from "./config";
 import { createDefaultFeedbackPromptHandler } from "./default-feedback-prompt-handler";
 import * as feedbackLib from "./feedback";
+import { getAuthToken } from "./prompt-storage";
 import {
   FeedbackPromptCompletionHandler,
   parsePromptMessage,
@@ -27,7 +28,6 @@ import type {
   TrackedEvent,
   User,
 } from "./types";
-import { getAuthToken } from "./prompt-storage";
 
 async function request(url: string, body: any) {
   return fetch(url, {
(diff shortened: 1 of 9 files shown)


roncohen:
how about also skipping calls to our ingest if we just did those calls? (non-spa apps)

pavkam:
> how about also skipping calls to our ingest if we just did those calls? (non-spa apps)

This one avoids one call to Ingest and one to Ably. One more call remains - is the `prompt-init`. I guess we can avoid it as well indeed. Let me see what I can do.

src/main.ts
```diff
@@ -324,38 +325,45 @@ export default function main() {
...
+
     // while initializing, consider the channel active
     liveSatisfactionActive = true;
     try {
-      const res = await request(`${getUrl()}/feedback/prompting-init`, {
-        userId,
-      });
+      if (!channel) {
+        const res = await request(`${getUrl()}/feedback/prompting-init`, {
+          userId,
```
roncohen, line 336: we use this in the app to show "initialized status" right?
Screenshot 2023-12-12 at 13 12 14

do we need some kind of expiry for that to ensure an reasonably up-to-date initialized time?

pavkam: It is set to expire when the token's TTL expires (that is by default an hour). But even then, the initialization check only cares about at least one connection attempt, which is a prerequisite for anything being cached.
In either case, we're covered.

roncohen: who controls the TTL? We currently say "Last client activity" which could then be confusing since it might be up to an hour delayed?

pavkam: I see what you mean. I thought you were referring to the indicator of "installed". The last activity is based on a `prompting-auth` call. So if we want to cache this across pages we'd need to sacrifice a bit of precision.

We control the TTL in ingest (default is `60min`). We can set it to lower values, but that would increase the number of refresh requests from clients. I think `60m` is reasonable.

roncohen: got it. Maybe we can just adjust the display to say something more vague rather than the usual relative time? Like "Today" if it's within 24h and then regular relative time after that? the green color is still nice. I'd love to have that color on there still if it's "today"

pavkam: Certainly can be done, add some level of ambiguity as an option to those time cells. I'll look into it.

test/usage.test.ts
```diff
@@ -299,6 +303,28 @@ describe("feedback prompting", () => {
     expect(closeAblySSEChannel).toBeCalledWith("fake_client");
   });
 
+  test("does not call tracking endpoint if token cached", async () => {
```
Munter, line 306: I can't make sense of the title of the test. It doesn't seem like we are testing that anything is specifically not called. There is no usage of nock for request interception/assertion and there is no assertion to a spy not being called. Is there something missing here?

src/prompt-storage.ts
```diff
@@ -15,3 +19,33 @@ export const checkPromptMessageCompleted = (
...
+
+export const getAuthToken = (userId: string) => {
+  const val = Cookies.get(`bucket-token-${userId}`);
+  if (!val) {
+    return undefined;
+  }
+
+  const [channel, token] = val.split(":");
+  if (!channel?.length || !token?.length) {
+    throw new Error(`Invalid token: ${val}`);
```
Munter, line 44: Do we want to throw as part of this interface? We run the risk of not catching it somewhere and getting into a non-working state in the client.

Also, If we have a known non-working cookie payload, maybe we should clear the cookie at this point

src/sse.ts
```diff
@@ -75,6 +77,12 @@ export class AblySSEChannel {
   }
 
   private async refreshToken() {
+    const cached = getAuthToken(this.userId);
```
Munter, line 80: This can throw, but we have no try/catch

src/main.ts
```diff
@@ -324,38 +325,45 @@ export default function main() {
 
     userId = resolveUser(userId);
 
+    const existingAuth = getAuthToken(userId);
```
Munter, line 328: This can throw, but we have no try/catch

Munter (changes requested):
I think we should change the `getAuthToken` interface to never throw and instead just return `undefined` in the case of an invalid cookie payload

src/prompt-storage.ts
pavkam: You're right, I wanted to be through since the cookie would be set by us and such an incorrect value would be an error on our side. But, I think it makes sense to simply ignore it and return `undefined` which would be equivalent to - `no cookie`.

src/main.ts
pavkam: Can't throw anymore.

src/sse.ts
pavkam: can't throw anymore

test/usage.test.ts
pavkam: This would fail exactly because there is no nock expecting a call. Also the ``` vi.mocked(getAuthToken).mockReturnValue({
      channel: "super-channel",
      token: "something",
    });
``` guarantees that `token`'s value `something` can only come from the storage as there is no other place that can get you a token.

Comment (Munter) [reply test/usage.test.ts:306]: We're actually not setting up nock anywhere in this test file, except for in individual other tests. As far as I can see, there is no nock instance active which could detect an unwanted request. If there is actually one, then it must be leaked from another test, which would be problematic.

Comment (pavkam) [reply test/usage.test.ts:306]: I probably explained incorrectly -- this test would fail if a nock is installed or not, if the code tries to call the API to obtain an auth token. So the test effectively tests that no such call is ever made. I can include a nock to explicitly show that.

Comment (pavkam) [reply test/usage.test.ts]: @Munter, done, added an explicit check

File: test/usage.test.ts [hunk 6]
```diff
@@ -299,6 +303,38 @@ describe("feedback prompting", () => {
...
+    expect(openAblySSEChannel).toBeCalledWith(
+      `${TRACKING_HOST}/${KEY}/feedback/prompting-auth`,
+      "foo",
+      "super-channel",
+      expect.anything(),
+      expect.anything(),
+    );
+
+    expect(n.isDone()).toBe(false);
+    n.done;
```
Munter, line 335:
```suggestion
    n.done();
```

Munter (approved):

PR #93, opened 2024-01-11
fix: auth cookie serialization/deserialization

matus-vacula:
This PR fixes an issue with cached token serialization and adds preventive fix in case a token is manually modified or possibly invalid due to server-side credential changes.

### DIFF
--- a/src/prompt-storage.ts
+++ b/src/prompt-storage.ts
@@ -26,7 +26,7 @@ export const rememberAuthToken = (
   token: string,
   expiresAt: Date,
 ) => {
-  Cookies.set(`bucket-token-${userId}`, JSON.stringify([channel,token]), {
+  Cookies.set(`bucket-token-${userId}`, JSON.stringify([channel, token]), {
     expires: expiresAt,
     sameSite: "strict",
     secure: true,
--- a/src/prompt-storage.ts
+++ b/src/prompt-storage.ts
@@ -26,7 +26,7 @@ export const rememberAuthToken = (
   token: string,
   expiresAt: Date,
 ) => {
-  Cookies.set(`bucket-token-${userId}`, JSON.stringify([channel, token]), {
+  Cookies.set(`bucket-token-${userId}`, JSON.stringify({ channel, token }), {
     expires: expiresAt,
     sameSite: "strict",
     secure: true,
@@ -40,7 +40,10 @@ export const getAuthToken = (userId: string) => {
   }
 
   try {
-    const [channel, token] = JSON.parse(val);
+    const { channel, token } = JSON.parse(val) as {
+      channel: string;
+      token: string;
+    };
     if (!channel?.length || !token?.length) {
       return undefined;
     }
--- a/src/prompt-storage.ts
+++ b/src/prompt-storage.ts
@@ -26,7 +26,7 @@ export const rememberAuthToken = (
   token: string,
   expiresAt: Date,
 ) => {
-  Cookies.set(`bucket-token-${userId}`, `${channel}:${token}`, {
+  Cookies.set(`bucket-token-${userId}`, JSON.stringify([channel,token]), {
     expires: expiresAt,
     sameSite: "strict",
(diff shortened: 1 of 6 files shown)


test/prompt-storage.test.ts
```diff
@@ -44,17 +44,21 @@ describe("prompt-storage", () => {
...
     rememberAuthToken("user", "channel", "token", new Date("2021-01-01"));
 
-    expect(spy).toHaveBeenCalledWith("bucket-token-user", "channel:token", {
-      expires: new Date("2021-01-01"),
-      sameSite: "strict",
-      secure: true,
-    });
+    expect(spy).toHaveBeenCalledWith(
+      "bucket-token-user",
+      '["channel","token"]',
```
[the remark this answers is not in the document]
pavkam: 👍🏽

pavkam (approved):

matus-vacula:
Good from my side but cannot hit "Approve" since I opened PR :+1:

src/prompt-storage.ts
```diff
@@ -26,7 +26,7 @@ export const rememberAuthToken = (
   token: string,
   expiresAt: Date,
 ) => {
-  Cookies.set(`bucket-token-${userId}`, `${channel}:${token}`, {
+  Cookies.set(`bucket-token-${userId}`, JSON.stringify([channel, token]), {
```
roncohen, line 29: How about storing an actual object with keys to avoid any mistakes in ordering or similar?

Munter: Seconded

Munter (commented):
I notice that at n point are we actually testing a full end-to-end writing and reading of the cookie. We're only using mocks. Since we ran into problems during the development of this branch I think it would be prudent to add a test that doesn't actually mock the setting and reading of cookies. vitest has a jsdom enviromnent available, which I assume will be able to write and read cookies, including handling of cookie expiration etc

pavkam:
> I notice that at n point are we actually testing a full end-to-end writing and reading of the cookie. We're only using mocks. Since we ran into problems during the development of this branch I think it would be prudent to add a test that doesn't actually mock the setting and reading of cookies. vitest has a jsdom enviromnent available, which I assume will be able to write and read cookies, including handling of cookie expiration etc

That would be testing the functionality of the `js-cookie` library in that case. Not sure we need to go that far. As long as we're testing the calls into it I think we're good. But if you think that's a good strategy, I can add a deeper test.

src/prompt-storage.ts
pavkam: done

Munter:
> That would be testing the functionality of the js-cookie library in that case. Not sure we need to go that far. As long as we're testing the calls into it I think we're good. But if you think that's a good strategy, I can add a deeper test.

Unless we have a good explanation for why @matus-vacula hit a case where cookie readind was not unescaped correctly, and a guarantee that we cannot hit this case accidentally again, I do indeed think we should be testing the full flow, including the cookie library. We might want to upgrade the library at some point, and with no test to verify that functionality is intact, we might well regress here in the future

matus-vacula:
I think the escaping problem might have been from an incorrect linking of the SDK.

pavkam:
> > That would be testing the functionality of the js-cookie library in that case. Not sure we need to go that far. As long as we're testing the calls into it I think we're good. But if you think that's a good strategy, I can add a deeper test.
> 
> Unless we have a good explanation for why @matus-vacula hit a case where cookie readind was not unescaped correctly, and a guarantee that we cannot hit this case accidentally again, I do indeed think we should be testing the full flow, including the cookie library. We might want to upgrade the library at some point, and with no test to verify that functionality is intact, we might well regress here in the future

Right, I still think it's an overkill but I have basically added all kinds of tests on "real cookies". This will at least ensure the formatting stays consistent across versions of `js-cookie`

roncohen (approved):
tried sending a prompt locally and that works 👍

PR #94, opened 2024-01-12
Update most of dev dependancies to their latest versions

pavkam:
Updating dev dependencies only.


matus-vacula (approved):

Munter (approved):
Looks good. I also verified the styling still works; this is a thing that can sometimes break when updating dev tooling and versions don't entirely match up

PR #99, opened 2024-03-19
Support offsetting feedback dialog

rrpff:
image

Can use any CSS value, or a number as a pixel value.

### DIFF
--- a/FEEDBACK.md
+++ b/FEEDBACK.md
@@ -173,8 +173,8 @@ position: {
   type: "DIALOG",
   placement: "top-left" | "top-right" | "bottom-left" | "bottom-right"
   offset?: {
-    x?: string | number; // e.g. "-5rem", "10px" or 10 (pixels)
-    y?: string | number;
+    left?: string | number; // e.g. "-5rem", "10px" or 10 (pixels)
+    top?: string | number;
   }
 }
 ```
--- a/FEEDBACK.md
+++ b/FEEDBACK.md
@@ -173,8 +173,8 @@ position: {
   type: "DIALOG",
   placement: "top-left" | "top-right" | "bottom-left" | "bottom-right"
   offset?: {
-    x?: number;
-    y?: number;
+    x?: string | number; // e.g. "-5rem", "10px" or 10 (pixels)
+    y?: string | number;
   }
 }
 ```
--- a/FEEDBACK.md
+++ b/FEEDBACK.md
@@ -168,10 +168,14 @@ Using a dialog is a soft push for feedback. It lets the user continue their work
 
 The default feedback UI behaviour is a dialog placed in the bottom right corner of the viewport.
 
-```javascript
+```typescript
 position: {
   type: "DIALOG",
   placement: "top-left" | "top-right" | "bottom-left" | "bottom-right"
+  offset?: {
+    x?: number;
+    y?: number;
+  }
 }
 ```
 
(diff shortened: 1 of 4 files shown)


FEEDBACK.md
```diff
@@ -168,10 +168,14 @@ Using a dialog is a soft push for feedback. It lets the user continue their work
...
 The default feedback UI behaviour is a dialog placed in the bottom right corner of the viewport.
 
-```javascript
+```typescript
 position: {
   type: "DIALOG",
   placement: "top-left" | "top-right" | "bottom-left" | "bottom-right"
+  offset?: {
+    x?: string | number; // e.g. "-5rem", "10px" or 10 (pixels)
+    y?: string | number;
```
Munter, line 177: These don't match the implementation

src/feedback/FeedbackDialog.tsx
```diff
@@ -69,18 +70,33 @@ export const FeedbackDialog: FunctionComponent<FeedbackDialogProps> = ({
...
   if (position.type === "DIALOG") {
+    const offsetLeft = parseOffset(position.offset?.left);
+    const offsetTop = parseOffset(position.offset?.top);
+
     switch (position.placement) {
       case "top-left":
-        unanchoredPosition = { top: "1rem", left: "1rem" };
+        unanchoredPosition = {
+          top: `calc(1rem + ${offsetTop})`,
+          left: `calc(1rem + ${offsetLeft})`,
```
Munter, line 80: I'm not sure I like this. I think I would prefer to have the user defined offset be unmodified and us just having an `1rem` default that the user can replace

src/feedback/types.ts
```diff
@@ -6,9 +6,14 @@ export type FeedbackPlacement =
   | "top-right"
   | "top-left";
 
+export type Offset = {
+  left?: string | number;
+  top?: string | number;
```
Munter, line 11: `top` and `left` are ambiguous, since this is in a positioning context where we already place the box in one of four corners. I think we either need to go more abstract, like the `x` and `y` from the docs or `horisontal`/ `vertical`

FEEDBACK.md
rrpff: They do now

src/feedback/FeedbackDialog.tsx
rrpff: Hmm, I was concerned it was pretty prone to human error. I revised the implementation to treat the offset as 'distance from the specified placement' rather than 'arbitrary CSS modifier' and that's simplified things

src/feedback/types.ts
rrpff: Felt the opposite before actually, I've seen `y` mean 'from the top' or 'from the bottom' in different contexts. I've reverted back to x/y though because of the implementation change though

src/feedback/types.ts
```diff
@@ -6,9 +6,14 @@ export type FeedbackPlacement =
   | "top-right"
   | "top-left";
 
+export type Offset = {
+  x?: string | number;
+  y?: string | number;
```
Munter, line 11:
```suggestion
  /**
   * Offset from the nearest horizontal screen edge after placement is resolved
   */
  x?: string | number;
  /**
   * Offset from the nearest vertical screen edge after placement is resolved
   */
  y?: string | number;
```

Munter (approved):
Looks good. I suggest a few comments to help people using autocomplete to discover the options in an editor

Swiftwork (approved):

roncohen:
Let’s get this in?

PR #108, opened 2024-04-13
Feature flagging

roncohen:


src/flags.ts
```diff
@@ -0,0 +1,266 @@
...
+      }
+    } else {
+      result[key] = obj[key];
+    }
+  }
+  return result;
+}
+
+const localStorageCacheKey = `__bucket_flags`;
+interface cacheEntry {
```
rrpff, line 133: This cache stuff is fairly dense, don't think it'd hurt to make it more generic (looks it already, could just use a key other than `flags`) and split out into a separate file with some unit tests

rrpff (approved):
Looks solid 👍

src/main.ts
```diff
@@ -558,6 +562,64 @@ export default function main() {
...
+   *                argument.
+   * @param fallbackFlags Optional array of flags to use if the request to Bucket fails.
+   * @param includeFlags Optional array of flags to include in the response automatically.
+   * @param timeoutMs Optional the timeout in milliseconds for the request to Bucket before using `fallbackFlags`. Default is 5000ms.
+   * @param staleWhileRevalidate Optional whether to return potentially stale flags while revalidating the flags in the background. Default is true.
+   */
+  async function getFeatureFlags({
+    context,
+    fallbackFlags = [],
+    includeFlags = [],
```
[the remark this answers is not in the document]
roncohen: good point, I'll just take it out.

src/flags.ts
roncohen: great idea!

PR #111, opened 2024-05-13
fix: send SDK version along with fetch requests

roncohen:
We're using a GET request to avoid CORS preflight requests for feature flags. This requires us to send the SDK version information along in a query string instead of the custom header we use for POST requests.

Updated the header/query param name to be lower-case. For headers it doesn't matter because headers are case-insensitive, but for the query param it's simpler to just have things be lower-case.

### DIFF
--- a/src/flags-fetch.ts
+++ b/src/flags-fetch.ts
@@ -1,8 +1,8 @@
 import fetch from "cross-fetch";
 
+import { SDK_VERSION, SDK_VERSION_HEADER_NAME } from "./config";
 import { Flags } from "./flags";
 import { FlagCache, isObject, validateFlags } from "./flags-cache";
-import { SDK_VERSION, SDK_VERSION_HEADER_NAME } from "./config";
 
 // Deep merge two objects.
 export function mergeDeep(
--- a/src/flags-fetch.ts
+++ b/src/flags-fetch.ts
@@ -81,10 +81,10 @@ export async function fetchFlags(url: string, timeoutMs: number) {
       const id = setTimeout(() => controller.abort(), timeoutMs);
 
       // add SDK version to the query params
-      const separator = url.includes("?") ? "&" : "?";
-      const newUrl = `${url}${separator}${SDK_VERSION_HEADER_NAME}=${SDK_VERSION}`;
+      const urlObj = new URL(url);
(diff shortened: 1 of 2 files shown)


src/flags-fetch.ts
```diff
@@ -78,7 +79,12 @@ export async function fetchFlags(url: string, timeoutMs: number) {
     try {
       const controller = new AbortController();
       const id = setTimeout(() => controller.abort(), timeoutMs);
-      const res = await fetch(url, {
+
+      // add SDK version to the query params
+      const separator = url.includes("?") ? "&" : "?";
+      const newUrl = `${url}${separator}${SDK_VERSION_HEADER_NAME}=${SDK_VERSION}`;
+
+      const res = await fetch(newUrl, {
```
Munter, line 87: String concatenation for urls always leads to poblems. Let's use the url object for safe manipulation instead.

```suggestion
      const urlObj = new URL(url);
      urlObj.set(SDK_VERSION_HEADER_NAME, SDK_VERSION);

      const res = await fetch(urlObj, {
```

Munter (approved):
Looks ok. I assume ingest needs to support both transport mechanisms indefinitely since old versions may never disappear

src/flags-fetch.ts
roncohen: nice!

roncohen:
> Looks ok. I assume ingest needs to support both transport mechanisms indefinitely since old versions may never disappear

The POST requests are still using the header actually. That felt like the right way to go about it, but because of the GET we'd need to do this hack. We could of course streamline it, but the query string with the POST also felt a bit weird when there's both headers and potentially a post body for that sort of thing. Thoughts?

Munter:
@roncohen 

> The POST requests are still using the header actually. The header felt like the right way to go about it in general

I agree. It's better to keep it as a sidecar when it doesn't negatively affect anything.

I'm interested though; If we could have the tracking requests go through without a CORS preflight, this might also help our tracking speed and consistency. It might actually be worth it. And in POST requests, despite it not being as neat to send query parameters, it's still completely valid

roncohen:
True. There are some differences around the fact that the GET request for flags are "idempotent" and could be cached, but we wouldn't want the tracking requests to be cached or retried. I'll file it under something to consider for the next big refactor ;)

PR #113, opened 2024-05-17
Monorepo

rrpff:


roncohen (approved):
"FEEDBACK.md" should probably go to tracking-sdk until we have react specific instructions for that?

Otherwise LGTM

.github/workflows/publish.yml
```diff
@@ -19,8 +20,9 @@ jobs:
           scope: "@bucketco"
       - name: Install dependencies
         run: yarn install --immutable
-      - name: Build SDK
+      - name: Build packages
         run: yarn build
-      - run: npm publish
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
roncohen, line 26: just checking, but this'll only publish if there's been a version change, right? I guess any attempt at publishing will fail if there's no new version anyway.

rrpff: Yep exactly. I just figured I'd disable until I make a version PR later today so I can iron out any possible issues then

PR #115, opened 2024-05-30
Fix behaviour around failed FF responses

rrpff:
The behaviour on a failed response is now to:
- try and extend the expiration of the last cached response if possible
- cache the failure if not
- but always refetch on load if the cached response is a failure

### DIFF
--- a/packages/tracking-sdk/src/flags-fetch.ts
+++ b/packages/tracking-sdk/src/flags-fetch.ts
@@ -160,21 +160,22 @@ export async function getFlags({
   const url = `${apiBaseUrl}/flags/evaluate?` + params.toString();
   const cachedItem = cache.get(url);
 
-  if (cachedItem && cachedItem.success) {
-    if (!cachedItem.stale) {
-      return cachedItem.flags;
-    }
+  if (!cachedItem) {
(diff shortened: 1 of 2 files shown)


packages/tracking-sdk/src/flags-fetch.ts
```diff
@@ -97,9 +97,17 @@ export async function fetchFlags(url: string, timeoutMs: number) {
...
       console.error("fetching flags: ", e);
     } finally {
-      // we cache even if the request failed to avoid having the UI
-      // wait and spam the API
-      cache.set(url, success, flags);
+      if (success) {
+        cache.set(url, success, flags);
+      } else if (cache.has(url)) {
+        // if there is a previous version, extend it's expireAt time
+        cache.extendExpiry(url);
```
roncohen, line 104: drive-by comment, but could this just be a `get` followed by a `set`?

roncohen:
could you rebase/merge main here?

matus-vacula (approved):
Can we move this forward? I still cannot see feature flags on prod

makwarth:
Can we get this shipped? I can't see any FFs on prod

makwarth:
Yay!

PR #136, opened 2024-06-18
Update react readme

Munter:
I noticed we were being a bit too magical in our documentation on how to use the react sdk.

- Added import statements
- Used `BucketProvider` as a name for the default export for more clarity


roncohen (approved):
nice! thanks @Munter

PR #138, opened 2024-06-18
feat(ReactSDK): include error states

roncohen:
Include error states and set `isLoading` to false anytime either an error happens or the load completes.


pavkam (approved):
Looks good

roncohen:
`getFeatureFlags` doesn't throw, so this isn't helpful.

PR #140, opened 2024-06-19
feat: allow multiple attempts to fail before caching a negative result

roncohen:
Requests for feature flags get cancelled if there's a page change around the same time as the flag request goes out. There's no reliable way to know if a request was cancelled or otherwise failed.

This PR changes the caching logic so we will not use a negative cached result until it's failed 3 times. If any of the first three requests succeed, that response will be used.

### DIFF
--- a/packages/tracking-sdk/src/flags-cache.ts
+++ b/packages/tracking-sdk/src/flags-cache.ts
@@ -11,6 +11,7 @@ interface cacheEntry {
   staleAt: number;
   success: boolean; // we also want to cache failures to avoid the UI waiting and spamming the API
   flags: Flags | undefined;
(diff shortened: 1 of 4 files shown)


pavkam (commented):
Too bad we can't use something like https://www.npmjs.com/package/cockatiel. That would be more flexible as we could cover more extraordinary conditions.

Also is there a way we can give control to the caller in case we can't fetch flags so they can do some sort of fallback?

roncohen:
> is there a way we can give control to the caller in case we can't fetch flags so they can do some sort of fallback?

There's `fallbackFlags` that you can pass. They will be used in case the requests fail.

pavkam (approved):

PR #144, opened 2024-06-24
Start sending 'check' events from SDK

pavkam:
BUC-2523

* Using a `Proxy` object for the value. Each time it is read, a custom function is called that triggers an event,
* Using a per-key rate limiter of 5 events per minute to avoid spamming the API,
* Added the `version` into the mix as we'll need that on API side,
* Fallback flags are also reported with no version attached, we can deal with them as want on the API side (maybe send a bool that it was unresolved, or just warn),
* Some cleaning was required related to the use of `nock` to make validation of calls tighter.


packages/tracking-sdk/src/main.ts
```diff
@@ -638,5 +702,6 @@ export default function main() {
     initLiveFeedback,
     // feature flags
     getFeatureFlags,
+    featureFlagEvent,
```
Munter, line 705: I don't think this should be a public interface

Munter (approved):
Looks good to me.

I don't think we should expose the concept of sending check events in the public interface though. It's very much a Bucket internal thing

roncohen:
@Munter I asked @pavkam to include it because it was thinking there could be situations where our Proxy implementation doesn't completely cover the use cases. However, smaller interfaces are better so I'm OK to remove it now and we can add it back in if it becomes necessary. Apologies for the back/forth @pavkam

roncohen:
@pavkam I've been wanting to rebase the ReactSDK work i've been doing on top of this in order to test that check events are being sent correctly. Lemme just give that a quick try, otherwise everything LGTM

pavkam:
> @Munter I asked @pavkam to include it because it was thinking there could be situations where our Proxy implementation doesn't completely cover the use cases. However, smaller interfaces are better so I'm OK to remove it now and we can add it back in if it becomes necessary. Apologies for the back/forth @pavkam

Will do!

pavkam:
> @pavkam I've been wanting to rebase the ReactSDK work i've been doing on top of this in order to test that check events are being sent correctly. Lemme just give that a quick try, otherwise everything LGTM

Will wait for you then

roncohen:
I tested it with the new React and it looks good. Nice work.

One thing was was apparent is that we need to make the rate limiter a bit smarter. Because of the way React works, you can have a ton of re-rendering going on. Each re-rendering can cause a "check" event to get emitted even though it's actually the same "check" (same flag key + context + result). Those could quickly saturate the rate limit causing check for new flags not to get transmitted. Can we make a rate limit of 1 event/minute but there's a separate ratelimit for each key/context combination?

pavkam:
> I tested it with the new React and it looks good. Nice work.
> 
> One thing was was apparent is that we need to make the rate limiter a bit smarter. Because of the way React works, you can have a ton of re-rendering going on. Each re-rendering can cause a "check" event to get emitted even though it's actually the same "check" (same flag key + context + result). Those could quickly saturate the rate limit causing check for new flags not to get transmitted. Can we make a rate limit of 1 event/minute but there's a separate ratelimit for each key/context combination?

Hm, good find. Will add the change. Thanks!

packages/tracking-sdk/src/proxy.ts
```diff
@@ -0,0 +1,21 @@
...
+      const val = target[prop as K];
+
+      if (val !== undefined) {
+        callback?.(prop as K, val);
+      }
+
+      return target[prop as K];
+    },
+    set(_target, prop, _value) {
+      throw new Error(
```
roncohen, line 16: We shouldn’t throw in other people code. This library already does it, but let’s not introduce new throws at least

packages/tracking-sdk/src/main.ts
```diff
@@ -598,13 +610,65 @@ export default function main() {
...
         return acc;
       }, {} as Flags);
     }
 
-    return flags;
+    return proxify(flags, rateLimited(FLAG_EVENTS_PER_MIN, sendCheckEvent));
+  }
+
+  function sendCheckEvent(_: keyof Flag, flag: Flag) {
+    void featureFlagEvent({
```
roncohen, line 623: This can cause unhandled promise rejections which can look scary. Let’s catch them and log a warning or similar.

packages/tracking-sdk/src/main.ts
```diff
@@ -598,13 +610,65 @@ export default function main() {
...
+      action: args.action,
+      flagKey: args.flagKey,
+      flagVersion: args.flagVersion,
+      evalContext: args.evalContext,
+      evalResult: args.evalResult,
+      evalRuleResults: args.evalRuleResults,
+      evalMissingFields: args.evalMissingFields,
+    };
+
+    const res = await postRequest(makeUrl("flags/events"), payload);
```
roncohen, line 667: This sends the publishable key as a query param. The pattern should be:
- always use the “Authorization” header
- except for GET requests

roncohen (approved):
Couple of more thoughts. Otherwise LGTM!

packages/tracking-sdk/src/main.ts
pavkam: That would add a pre-flight for all events we send - `user`, `company` and `track` events since they are POSTs

packages/tracking-sdk/src/proxy.ts
pavkam: I don't know if silently ignoring the new value is better, to be honest. And if we allow them to set the value then we might get invalid feature flags in our events. It's better to explicitly disallow this behavior.

packages/tracking-sdk/src/main.ts
roncohen: The ingest accepts POST body requests when they use the `Content-type: application/json` header. `Content-type: application/json` will trigger a preflight request regardless of the rest. So, we're already triggering pre-flight requests for those - and that's OK. There's no latency requirement for those, just for the feature flag requests.

Screenshot 2024-07-10 at 10 26 25

packages/tracking-sdk/src/main.ts
```diff
@@ -589,22 +601,81 @@ export default function main() {
...
+      flags = fallbackFlags.reduce((acc, flag) => {
         acc[flag.key] = flag;
         return acc;
       }, {} as Flags);
     }
 
-    return flags;
+    return proxify(
+      flags!,
+      rateLimited(FLAG_EVENTS_PER_MIN, res.url, sendCheckEvent),
```
roncohen, line 623: The flag key is not part of the `res.url`. So if you use two different flags within 1 minute, we'll only collect a check for one.

The check event should include:
- flagKey
- flagVersion: flag.version,
- evalResult
- context

and all of these should be in the key

packages/tracking-sdk/src/main.ts
pavkam: Updated the SDK to use the new style endpoints with the required params/headers

roncohen (approved):

packages/tracking-sdk/src/config.ts
```diff
@@ -1,8 +1,9 @@
 import { version } from "../package.json";
 
-export const API_HOST = "https://tracking.bucket.co";
+export const API_HOST = "https://front.bucket.co";
```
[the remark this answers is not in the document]
pavkam: done

pavkam:
Closing as this package is superseded by browser SDK which will incorporate all the additions.

PR #145, opened 2024-06-26
New ReactSDK API

roncohen:
Updated React SDK API. Many changes across the board. See the [README.md](https://github.com/bucketco/bucket-javascript-sdk/blob/new-api/packages/react-sdk/README.md) how to use it now.


Munter (commented):
This is comments only, since I won't be around to get this fully closed.

I've left some inline comments.

I am a bit worried about the huse single `useEffect` that runs on literally every change of anything. It resets the client every time, which seems wasteful as this will essentially log the user out, kill the ably connection etc. Further, it will likely send an unnessessary amount of user/company calls when no changes actually happened.

I think the user/company trakcking calls should be in each their own useEffect, and possibly avoid triggering f no changes to their specific model was made.

Further, the reset should probably only be called when unmounting the provider

roncohen:
thanks @Munter !

packages/react-sdk/src/index.tsx
```diff
@@ -29,39 +66,93 @@ const Context = createContext<BucketContext>({
...
+  // if user.id or attributes change, send new attributes to the servers
   useEffect(() => {
-    try {
-      const { publishableKey, flags: flagOptions, ...options } = config;
+    if (user?.id) {
+      const { id, ...attributes } = user;
+      // `user` calls bucket.reset() automatically when needed
+      bucket.user(String(id), attributes);
+    }
+  }, [canonicalJSON({ user })]);
```
Swiftwork, line 123: Rather than relying on `useEffect`, it would probably be more elegant to create `useCallback` functions that you expose in the context. So that `useUser` calls a `setUser` from the context. `useEffect` is a bit of a crutch and it's easy to rely on it but in my experience most React errors are somehow related to `useEffect` triggering too frequently or not enough. https://react.dev/learn/you-might-not-need-an-effect

packages/react-sdk/src/index.tsx
```diff
@@ -29,39 +66,93 @@ const Context = createContext<BucketContext>({
...
+    company: initialCompany,
+    otherContext: initialOtherContext,
+  });
+
+  const { user, company, otherContext } = flagContext;
+
+  // on mount
+  useEffect(() => () => bucket.init(config.publishableKey, config), []);
+  // on umount
+  useEffect(() => () => bucket.reset());
```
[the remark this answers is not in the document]
Swiftwork: Hmm, reading @Munter comment about opening and closing connections is interesting. Generally, we don't want to have no dependency array or an empty dependency array `[]` as that leads to the issue mentioned below about the effect triggering too frequently or not enough https://react.dev/learn/you-might-not-need-an-effect. If we want the init only to be run once we should probably add a `const firstRun = useRef(false)` that gets updated when init has been called. I would have to dig deeper to understand and make sure the `init` and `reset` are called the "correct" amount of times. Maybe a chat @roncohen could be helpful as well?

packages/react-sdk/src/index.tsx
```diff
@@ -29,39 +66,93 @@ const Context = createContext<BucketContext>({
...
+  const [flagContext, setFlagContext] = useState({
+    user: initialUser,
+    company: initialCompany,
+    otherContext: initialOtherContext,
+  });
+
+  const { user, company, otherContext } = flagContext;
+
+  // on mount
+  useEffect(() => () => bucket.init(config.publishableKey, config), []);
```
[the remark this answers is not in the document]
roncohen: true! fixed it. Thank!

packages/react-sdk/README.md
```diff
@@ -12,102 +12,221 @@ npm i @bucketco/react-sdk
...
+  fallbackFlags={["huddle"]} // if we're unable to fetch flags, these flags will be enabled.
 >
-  {/* ... */}
-</BucketProvider>;
+{/* children here are shown when loading finishes */}
+</BucketProvider>
 ```
 
-### Props
+- `publishableKey` is used to connect the provider to an _environment_ on Bucket. Find your `publishableKey` under `Activity` on https://app.bucket.co.
```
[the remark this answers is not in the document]
roncohen: We're changing the name :)

packages/react-sdk/README.md
```diff
@@ -12,102 +12,221 @@ npm i @bucketco/react-sdk
...
+<BucketProvider publishableKey={YOUR_PUBLISHABLE_KEY} /*...*/>
+  <LoadingBucket>
+   {/* children here are shown when loading finishes */}
+   </LoadingBucket>
+<BucketProvider>
+```
+
+### Options
+
+TODO: Describe options to `BucketProvider`
```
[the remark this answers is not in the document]
roncohen: thanks! removed. They are all documented above.

packages/react-sdk/src/index.tsx
roncohen: changed them to useCallback as you suggested. Please have another look!

roncohen:
thanks @Swiftwork ! mind having another look? updated all the hooks including a rename of `useUser` etc. to go under `useUpdateContext`.

roncohen:
this example doesn't make much sense, but hopefully illustrative enough: https://github.com/bucketco/bucket-javascript-sdk/blob/new-api/packages/react-sdk/README.md#useupdatecontext

ideas are very welcome.

packages/react-sdk/src/index.tsx
```diff
@@ -123,6 +130,9 @@ export function BucketProvider({
       const { id, ...attributes } = user;
       // `user` calls bucket.reset() automatically when needed
       bucket.user(String(id), attributes);
+    } else {
+      // logout
+      bucket.reset();
     }
   }, [canonicalJSON({ user })]);
```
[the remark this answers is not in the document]
roncohen: fixed, thanks!

packages/react-sdk/src/index.tsx
```diff
@@ -9,19 +10,58 @@ import canonicalJSON from "canonical-json";
...
+
+  sendFeedback: (opts: Omit<Feedback, "userId" | "companyId">) => void;
+  requestFeedback: (
+    opts: Omit<RequestFeedbackOptions, "userId" | "companyId">,
+  ) => void;
+
+  track: (eventName: string, attributes?: Record<string, any>) => void;
 };
 
 const Context = createContext<BucketContext>({
```
Swiftwork, line 67: I would consider calling this `BucketContext` instead and the type `BucketContextState` or similar.

packages/react-sdk/src/index.tsx
```diff
@@ -30,108 +70,327 @@ const Context = createContext<BucketContext>({
...
-  const flag = flags.flags[key];
+export function useUpdateContext() {
+  const {
+    updateUser,
+    updateCompany,
+    updateOtherContext,
+    flags: { isLoading },
+  } = useContext<BucketContext>(Context);
+  return { updateUser, updateCompany, updateOtherContext, isLoading };
+}
```
Swiftwork, line 342: I would consider naming this just `useBucket` and include all methods here rather than splitting them up into their own hooks. `useFlag` hooks can stay as they alter the return in convenient ways but the others don't add much value being isolated and could simply be.

```ts
const {
  track,
  requestFeedback,
  sendFeedback,
  updateUser,
  updateCompany,
  updateOtherContext,
} = useBucket();
```

Swiftwork (approved):

packages/react-sdk/src/index.tsx
roncohen: don't you think it's either/or? weird that `useFlag` is outside `useBucket` ?

packages/react-sdk/src/index.tsx
roncohen: since they are not exported and everything in this file is "bucket" i went for `ProviderContext` and `ProviderContextType`. I assume you commented because `context` could be conflated with flag context etc. ?

Swiftwork: I had a chat with @roncohen and we both agreed there were benefits and detriments to grouping it under one hook. Ultimately we decided to stick with the current approach and to test it in BETA first to get a feel for it.

packages/react-sdk/src/index.tsx
Swiftwork: I had a chat with @roncohen and we both agreed there were benefits and detriments to grouping it under one hook. Ultimately we decided to stick with the current approach and to test it in BETA first to get a feel for it.

PR #147, opened 2024-07-03
Node-sdk

pavkam:
Initial implementation of the `node-sdk`.

Please look at the code/API perfectly and give me all opinions and change requests.


packages/node-sdk/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+
+See [types](./src/types.ts) for more details on the supported types.
+
+You can then then use the `boundClient` to perform all the supported functionality:
+
+```ts
+// update the user with Bucket (await it optional)
+await boundClient.updateUser();
```
Munter, line 66: This example doesn't make a lot of sense. What would the effect of calling this function without any arguments be?

packages/node-sdk/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+// update the associated company with Bucket
+await boundClient.updateCompany();
```
Munter, line 69: This example doesn't make a lot of sense. What would the effect of calling this function without any arguments be?

packages/node-sdk/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export type { Client } from "./client";
```
Munter, line 1: Are we only exporting the type here? Seems like we should export the actual client.

I think we should name it `BucketClient` in the public interface. `Client` is quite overloaded and might result in forcing the consumer to change the import names

packages/node-sdk/README.md
```diff
@@ -0,0 +1,170 @@
...
+`withUsr()`, `withCompany()` and `withCompanyContext()` methods. Each of
+these three methods returns a new instance of `Client` class which includes
+the applied user, company or context:
+
+```ts
+// configure the client
+const boundClient = client
+  .withUser({ id: "john_doe" name: "John Doe" })
+  .withCompany({ id: "acme_inc", name "Acme Inc."})
+  .withCustomContext({ additional: "context", number: [1,2,3] })
```
Munter, line 56: Do these chains result in a new immutable instance for each call, or does the first one create a new instance which is then mutated by the remaining calls?

Only could easily expect user, company and context to be available in server middlewares at different points, so it won't be possible to set up this chain once and for all in a single place. Rather these pieces of information will have to be added piecemeal. If a new immutable instance is created each time, the client reference will have to be re-assigned in each place. It might be relevant to show an example of this.

If I wanted to check a flag without any user, company or custom context, would I just access the global client instance? I'm thinking of when we presumably add always on/off rules per environment. These could be evaluated fine without any additional context other than the secret key

packages/node-sdk/README.md
```diff
@@ -0,0 +1,170 @@
...
+client.withUser({ userId: await sha256("john_doe"), ... });
+```
+
+### Custom Attributes & Context
+
+You can pass attributes as part of the object literal passed to the `withUser()`,
+`withCompany()` and `trackFeatureUsage()` methods. Attributes cannot be nested
+(multiple levels) and must be either strings, integers or booleans.
+
+Some attributes are used by Bucket.co and manged separately, such as:
```
Munter, line 127:
```suggestion
Some attributes are used by Bucket.co to improve the UI, and are recommended to provide for easier navigation:
```

packages/node-sdk/README.md
```diff
@@ -0,0 +1,170 @@
...
+  })
+  .withCompany({
+    companyId: "0083663",
+    attributes: { name: "My SaaS Inc." },
+    context: { active: false },
+  });
+
+// send updates to Bucket.co with the new details.
+await boundClient.updateCompany();
+await boundClient.updateUser();
```
Munter, line 159: This seems like an unintuitive way of triggering the underlying user and company tracking events to bucket. You might well want to register different things in bucket than you'd want to provide as a feature flagging context.

Especially the `active: true/false` seems like something that is not relevant to feature flag checks at all, but it only relevant to updating entities in bucket. So I would expect the `active`-meta to be a part of the function that triggers these bucket HTTP API calls

packages/node-sdk/src/config.ts
```diff
@@ -0,0 +1,9 @@
+import { version } from "../package.json";
+
+export const API_HOST = "https://tracking.bucket.co";
+export const SDK_VERSION_HEADER_NAME = "bucket-sdk-version";
+export const SDK_VERSION = version;
```
Munter, line 5: It might actually also help us to send the sdk name as an HTTP header, so we can break down usage even further. Right now we can just see a version, but won't know if it came from js, react, node or whatever.

Munter (commented):
First pass, partial review of what I had time for

packages/node-sdk/src/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+ * Assert that the given condition is `true`.
+ *
+ * @param condition - The condition to check.
+ * @param message - The error message to throw.
+ **/
+export function ok(condition: boolean, message: string): asserts condition {
+  if (!condition) {
+    throw new Error(message);
+  }
+}
```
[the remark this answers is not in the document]
pavkam: Discussed, and agreed to maintain our version.

packages/node-sdk/src/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (!target[key]) Object.assign(target, { [key]: {} });
+        mergeDeep(target[key], source[key]);
+      } else {
+        Object.assign(target, { [key]: source[key] });
+      }
+    }
+  }
+
+  return mergeDeep(target, ...sources);
+}
```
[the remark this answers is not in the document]
pavkam: Removed the deep merge. We don't need it

packages/node-sdk/src/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    if (events.length >= eventsPerMinute) {
+      return;
+    }
+
+    events.push(now);
+    return func(...funcArgs);
+  };
+}
```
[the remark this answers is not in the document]
pavkam: Decided to keep our version as it's simple enough

packages/node-sdk/README.md
pavkam: We discussed and came up with changes.

packages/node-sdk/src/config.ts
pavkam: added the name of the SDK

roncohen (commented):
Initial pass

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+Node.js, JavaScriptS/Typescript feature flag and tracking client for [Bucket.co](https://bucket.co).
+
+## Installation
+
+Install using `yarn` or `npm` with:
+
+> `yarn add -s @bucketco/node-sdk` or `npm install -s @bucketco/node-sdk`.
+
+Other languages than JavaScript/Typescript are currently not supported by an SDK.
+You can [use the HTTP API directly](https://docs.bucket.co/reference/http-tracking-api)
```
roncohen, line 12:
```suggestion
Other supported languages/frameworks are in the [Supported languages](https://docs.bucket.co/quickstart/supported-languages) documentation pages.

You can also [use the HTTP API directly](https://docs.bucket.co/reference/http-tracking-api)
```

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+Before the library can be used, you need to obtain a secret key from
+[Environment setting view](https://app.bucket.co/envs/{environment}/settings/app-environments)
+in **Bucket.co**.
+
+> [!CAUTION]
+> Secret keys should only be used by the `node-sdk` and not any other
+> SDKs provided by Bucket. Secret keys offer the users the ability to obtain
+> information that is often sensitive and thus should not be used in any
+> front-end application.
```
roncohen, line 24:
```suggestion
To get started you need to obtain a secret key from
[Environment setting view](https://app.bucket.co/envs/{environment}/settings/app-environments)
in **Bucket.co**.

> [!CAUTION]
> Secret keys are meant for use in server side SDKs only.
> Secret keys offer the users the ability to obtain
> information that is often sensitive and thus should not be used in
> client-side applications.
```

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+  // The interval to re-fetch feature flag definitions (optional).
+  // Default is 60 seconds.
+  refetchInterval?: number = 60000,
+  // The interval to re-fetch feature flags (optional). Default
+  // is 5 times the re-fetch interval.
+  staleWarningInterval?: number = 5 * 60000,
+  // The logger you can supply. By default no logging is performed.
+  logger?: Logger,
+  // The custom http client. By default the internal `fetchClient` is used.
+  httpClient?: HttpClient = fetchCient,
```
roncohen, line 160: missing `fallbackFlags`

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+  secretKey: "sec_prod_xxxxxxxxxxxxxxxxxxxxx",
+});
+
+// Initialize the client's feature flag retrieval loop. You must call this method
+// prior to any calls to `getFlags()`, otherwise an empty object will be returned.
+//
+// You can also supply your `fallbackFlags` to the `initialize()` method. These
+// fallback flags are used in the situation when the server-side flags are not
+// yet loaded or there are issues retrieving them.
+await client.initialize();
```
roncohen, line 45:
```suggestion
// Initialize the client and begin fetching flag definitions.
// You must call this method prior to any calls to `getFlags()`,
// otherwise an empty object will be returned.
//
// You can also supply your `fallbackFlags` to the `initialize()` method. These
// fallback flags are used in the situation when the server-side flags are not
// yet loaded or there are issues retrieving them.
// See "Initialization Options" below for more information.
await client.initialize();
```

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+// configure the client
+const boundClient = client
+  .withUser("john_doe", { attributes: { name: "John Doe" } })
+  .withCompany("acme_inc", { attributes: { name "Acme Inc."} })
+  .withCustomContext({ additional: "context", number: [1,2,3] })
+
+const fallbackFlags = { can_see_new_reports: { value: true } }
+
+// actively wait for the flags to be loaded
+await boundClient.initialize(fallbackFlags)
```
roncohen, line 63: Hm, I think fallbackFlags should be an argument to the `Client` constructor instead of to `initialize`?  Feels like it fits into the `Client` constructor with the other initialization arguments.

roncohen: i think the order here is a bit funky. We just showed how to bind the client and here comes the initialization example to show how to use the fallbackFlags. Let's make `fallbackFlags` part of the Client constructor and remove this example. I want the getting started examples to be as simple as possible so the `Client` creation example you have above that just takes the secretKey is good. Let's just explain the fallbackFlag in "Initialization Options" below.

roncohen (commented):
round two! :)

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+automatically update the given user/company `Last seen` property on Bucket side.
+You can control if `Last seen` should be updated when the events are sent by setting
+`context.active = false` to avoid updating last seen. This is often useful if you
+have a background job that goes through a set of companies just to update their.
+
+Example:
+
+```ts
+// create a new client initialized with an user and a company.
+client.trackUser("188762", {
```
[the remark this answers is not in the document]
pavkam: That was the initial implementation but after discussions with @Munter we settled on a manual approach. 

Otherwise I'd have to introduce 3 "bound classes": `bound-to-user`, `bound-to-company` and `bound-to-both`. Not a biggie but annoying enough.

packages/node-sdk/README.md
```diff
@@ -0,0 +1,223 @@
...
+client.trackcompany("updated_company_id", {
+  attributes: { status: "active", plan: "trial" },
+  meta: { active: false },
+});
+```
+
+To generate feature tracking `event`s:
+
+```ts
+// this simply sends an event to Bucket, not associated with any company or user.
```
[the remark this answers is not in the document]
pavkam: see comment below

packages/node-sdk/README.md
pavkam: Hm, so we won't be able to switch over to new fallback flags when we want? We would only be able to create a new instance of the client when we need new fallback flags.

roncohen: Was the idea that you can call initialize again if you want to change the fallback flags? I’d expect you’d be unable to call initialize more than once so that would be a surprising recommendation to me. I think it’s fine to require folks to create a new client if they want to switch the fallback flags.

packages/node-sdk/README.md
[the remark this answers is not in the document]
roncohen: What’s the point of “withUser” if you must still pass userId to track?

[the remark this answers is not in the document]
roncohen: apologies, I'm mixing things up. `trackFeatureUsage` doesn't take a `userId`. So nevermind 👍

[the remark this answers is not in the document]
pavkam: No, you were correct, we require user/company explicitly supplied for any `trackXXXX` method. I will make the changes to use bound clients if that's the decision. I wish @Munter was here since he had the opposite view. Would have liked to have all opinions on the table.

packages/node-sdk/src/config.ts
```diff
@@ -0,0 +1,11 @@
+import { version } from "../package.json";
+
+export const API_HOST = "https://tracking.bucket.co";
+export const SDK_VERSION_HEADER_NAME = "bucket-sdk-version";
+export const SDK_VERSION = `node/${version}`;
```
roncohen, line 5: nit, but let's call this `node-sdk`, so it's `node-sdk/${version}` to be sure.

packages/node-sdk/example/README.md
```diff
@@ -0,0 +1,8 @@
+# Bucket Node-SDK Express example
+
+## Running
+
+```
+yarn install
+yarn dev
```
roncohen, line 7:
```suggestion
EXPORT BUCKET_SECRET_KEY=<secretKey> yarn start
```

packages/node-sdk/example/bucket.ts
```diff
@@ -0,0 +1,13 @@
+import { BucketClient } from "@bucketco/node-sdk/src";
+
+export const bucket = new BucketClient({
+  secretKey: process.env.BUCKET_SECRET_KEY || "sec_prod_12345678901234567890",
```
roncohen, line 4: Let's remove the fallback? Maybe too late, but maybe it'll remove some false security warnings from scanners. I've added a suggestion to the README instead.

packages/node-sdk/example/README.md
```diff
@@ -0,0 +1,8 @@
+# Bucket Node-SDK Express example
+
+## Running
+
```
roncohen, line 4:
```suggestion
+ Create an app on bucket.co and take a note of the secret key which is found under "Settings" -> "Environments"
```

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+   **/
+  private async get<TResponse>(path: string) {
+    ok(typeof path === "string" && path.length > 0, "path must be a string");
+
+    try {
+      const response = await this._shared.httpClient.get<
+        TResponse & { success: boolean }
+      >(`${this._shared.host}/${path}`, this._shared.headers);
+
+      this._shared.logger?.debug(`get request to "${path}"`, response);
```
roncohen, line 184: I got this when using an invalid secret key:
```
yarn start
Server is running on port 3000
[Bucket] get request to "flags" { status: 403, body: undefined }
[Bucket] received undefined value from function
Bucket initialized
```

I think the `body` is getting lost? We normally give a good error message in the ingest.

packages/node-sdk/example/app.ts
```diff
@@ -0,0 +1,73 @@
...
+});
+
+app.post("/todos", (req, res) => {
+  const { todo } = req.body;
+  if (typeof todo !== "string") {
+    return res.status(400).json({ error: "Invalid todo" });
+  }
+
+  todos.push(todo);
+
```
roncohen, line 58: let's add a "create-todos" flags which gates the creation of flags? There's only one flag atm. and the functionality doesn't make that much sense.

roncohen: and maybe one for "delete-todos" too?

packages/node-sdk/example/app.ts
```diff
@@ -0,0 +1,73 @@
...
+        role: req.headers["x-bucket-is-admin"] ? "admin" : "user",
+      },
+    })
+    .withCompany(req.headers["x-bucket-company-id"] as string, {
+      attributes: {
+        betaUser: !!req.headers["x-bucket-company-beta-user"],
+      },
+    });
+
+  res.locals.bucketUser = bucketUser;
```
roncohen, line 30: Hmm, not sure what's going on here :/ 

Screenshot 2024-07-22 at 11 14 38

packages/node-sdk/README.md
```diff
@@ -0,0 +1,231 @@
...
+  .withUser("your_user_id", {
+    attributes: { longTimeUser: true, payingCustomer: false },
+  })
+  .withCompany("company_id");
+
+// track the user (send a `user` event to Bucket).
+await boundClient.trackUser();
+
+// register the user as being part of a given company
+boundClient.trackCompany();
```
roncohen, line 112: Maybe you and @Munter were already back and forth on this, but did you consider removing the `trackCompany` / `trackUser` methods and instead only relying on `withCompany` / `withUser` ? This is a feature flagging SDK and whenever you do `withCompany` and give it some attributes and then go on to evaluate some flags, those attributes need to be transmitted to our servers in order for the views/tables there to be accurate (who has access to what). So, I think it makes sense for `withCompany` to automatically call `trackCompany` at the least. And then we should ask ourselves if `trackCompany` and `trackUser` is even useful as separate methods.

packages/node-sdk/example/app.ts
```diff
@@ -0,0 +1,73 @@
...
+
+app.get("/", (_req, res) => {
+  res.locals.bucketUser.trackFeatureUsage("Front Page Viewed");
+  res.send("Hello World");
+});
+
+app.get("/todos", async (_req, res) => {
+  const flags = res.locals.bucketUser.getFlags<AppFlags>();
+
+  if (flags["show-todos"]) {
```
roncohen, line 44: this is actually wrong:
Screenshot 2024-07-22 at 11 22 54

for the `getFlags` equivalent in the ReactSDK we went with a simple `Record<string, boolean>` after feedback from Lasse/Peter: https://github.com/bucketco/bucket-javascript-sdk/blob/main/packages/react-sdk/src/index.tsx#L286-L314

Since there's no `loading` state for this call, i think it makes sense here too. I initially made it more complicated, but that was a premature abstraction and it's not useful at the moment.

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+    return readNotifyProxy(
+      evaluatedFlags,
+      rateLimited(FLAG_EVENTS_PER_MIN, async (_: string, res: Flag) => {
+        await this.sendFeatureFlagEvent({
+          action: "check",
+          flagKey: res.key,
+          flagVersion: res.version,
+          evalResult: res.value,
```
roncohen, line 602: we should also be sending the context along in these `check` events IMO

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+      this._shared.logger?.debug("evaluated flags", evaluatedFlags);
+    } else {
+      this._shared.logger?.warn(
+        "failed to use feature flag definitions, there are none cached yet. using fallback flags.",
+      );
+    }
+
+    return readNotifyProxy(
+      evaluatedFlags,
+      rateLimited(FLAG_EVENTS_PER_MIN, async (_: string, res: Flag) => {
```
roncohen, line 597: This should be using the same cache key calculation as in the tracking-sdk. E.g. send a check event per flag.key, context, version ... per minute

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+      const keyToVersionMap = new Map<string, number>(
+        flagDefinitions.flags.map((f) => [f.key, f.version]),
+      );
+
+      const evaluated = flagDefinitions.flags.map((flag) =>
+        evaluateFlag({ context: mergedContext, flag }),
+      );
+
+      evaluated.forEach(async (res) => {
+        await this.sendFeatureFlagEvent({
```
roncohen, line 563: this should probably also be rateLimited in the same way that "check" events are?

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+   *
+   * @returns A new client with the context set.
+   * @throws An error if the context is not an object or the options are invalid.
+   * @remarks
+   * If replace is true, the context will replace the current context, otherwise it will be merged.
+   * The new context will take precedence over the old context.
+   **/
+  public withOtherContext(
+    context: Record<string, any>,
+    opts?: { replace?: boolean },
```
roncohen, line 367: I would get rid of this `replace` boolean. If you need a "merge" operation it's easy to just do

```
client.withOtherContext({...client.otherContext, newKey: newValue})
```

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+   *
+   * @returns A new client with the company set.
+   * @throws An error if the company ID is not a string or the options are invalid.
+   * @remarks
+   * If the company ID is the same as the current company, the attributes will be merged, and
+   * the new attributes will take precedence.
+   **/
+  public withCompany(
+    companyId: string,
+    opts?: { attributes?: Attributes },
```
roncohen, line 328: do you already have something in mind for the other options? If not, let's make this simpler and just let is take the `attributes?: Attributes` as a second parameter without the `opts`. If we really want to add more options later, that's still possible.

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+   *
+   * @returns A new client with the user set.
+   * @throws An error if the user ID is not a string or the options are invalid.
+   * @remarks
+   * If the user ID is the same as the current company, the attributes will be merged, and
+   * the new attributes will take precedence.
+   **/
+  public withUser(
+    userId: string,
+    opts?: { attributes?: Attributes },
```
roncohen, line 289: do you already have something in mind for the other options? If not, let's make this simpler and just let is take the attributes?: Attributes as a second parameter without the opts. If we really want to add more options later, that's still possible.

packages/node-sdk/src/client.ts
```diff
@@ -0,0 +1,607 @@
...
+} from "./utils";
+
+/**
+ * The SDK client.
+ *
+ * @remarks
+ * The client is used to interact with the Bucket API.
+ * Use the client to track users, companies, events, and feature flag usage.
+ **/
+export class BucketClient {
```
roncohen, line 40: If you call `trackFeatureUsage` on the client directly, we'll throw. Did you look at trying to handle this through types instead? Something like this could potentially work. The idea is instead of calling `new BucketBaseClient` directly, you'll call `const client = BucketClient(...)`. It will return an interface without `trackFeatureUsage` so you'll need to do `withUser` first to get "access" to that method:

```typescript
type BucketUserClient = {
  withCompany(
    companyId: string,
    opts?: { attributes?: Attributes },
  ): BucketCompanyClient;

  trackFeatureUsage(event: string, opts?: TrackOptions): Promise<boolean>;
  trackUser(opts?: TrackOptions): Promise<boolean>;
  withOtherContext(
    context: Record<string, any>,
    opts?: { replace?: boolean },
  ): BucketUserClient;
  otherContext: Record<string, any> | undefined;
  user: { userId: string; attrs?: Attributes } | undefined;
} & BucketBaseClient;

export type BucketCompanyClient = {
  company: { companyId: string; attrs?: Attributes } | undefined;
  trackCompany(opts?: TrackOptions): Promise<boolean>;
} & BucketUserClient;

interface BucketClientI {
  withUser(
    userId: string,
    opts?: { attributes?: Attributes },
  ): BucketUserClient;
  initialize(): Promise<void>;
  getFlags<TFlagKey extends string = string>(): Readonly<
    Record<TFlagKey, Flag>
  >;
}

/**
 * The SDK client.
 *
 * @remarks
 * The client is used to interact with the Bucket API.
 * Use the client to track users, companies, events, and feature flag usage.
 **/
function BucketClient(options: ClientOptions): BucketClientI {
  return new BucketBaseClient(options);
}

class BucketBaseClient {
  private _shared: {
    logger?: Logger;
    host: string;
    httpClient: HttpClient;
    refetchInterval: number;
    staleWarningInterval: number;
[...]
```

packages/node-sdk/example/bucket.ts
pavkam: It just was a fake key, I was testing

packages/node-sdk/src/client.ts
pavkam: Yes, I do not parse the body in such cases, figured a good old `403` is enough. But I can add the message as well.

packages/node-sdk/example/bucket.ts
roncohen: yes of course, but the security scanners don't know that :)

packages/node-sdk/README.md
```diff
@@ -0,0 +1,231 @@
...
+boundClient.updateCompany();
+```
+
+If one needs to simply update a company's attributes on Bucket side,
+one calls `updateCompany` without supplying a user ID:
+
+```ts
+// either creates a new company on Bucket or updates an existing
+// one by supplying custom attributes
+client.withCompany("updated_company_id").updateCompany({
```
roncohen, line 121:
```suggestion
client.withCompany("company_id").updateCompany({
```

packages/node-sdk/README.md
```diff
@@ -0,0 +1,231 @@
...
+      some: "attribute",
+    },
+  });
+}
+```
+
+## Tracking
+
+Tracking allows sending `user`, `company` and `event` messages to Bucket.
+`user` events can be used to register or update an user's attributes
```
roncohen, line 92:
```suggestion
`user` events can be used to register or update a user's attributes
```

roncohen (approved):
Let’s get this in!

packages/node-sdk/README.md
```diff
@@ -0,0 +1,231 @@
...
+// configure the client
+const boundClient = client
+  .withUser("john_doe", { attributes: { name: "John Doe" } })
+  .withCompany("acme_inc", { attributes: { name "Acme Inc."} })
+  .withOtherContext({ additional: "context", number: [1,2,3] })
+
+const fallbackFlags = { can_see_new_reports: true  }
+
+// actively wait for the flags to be loaded
+await boundClient.initialize()
```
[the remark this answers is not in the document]
pavkam: It doesn't really matter when it's called, though the earlier - the better. Any subsequent calls will effectively refresh the flags. But let me correct the docs

PR #151, opened 2024-07-17
express example for node-sdk

roncohen:
Express.js example to show how to use the Node.js SDK


packages/node-sdk/example/bucket.ts
```diff
@@ -0,0 +1,8 @@
+import { BucketClient } from "../src";
+
+export const bucket = new BucketClient({
+  secretKey: process.env.BUCKET_SECRET_KEY || "",
+});
+
+await bucket.initialize();
```
pavkam, line 7: Does this `await` actually work at module level?

packages/node-sdk/example/bucket.ts
```diff
@@ -0,0 +1,8 @@
+import { BucketClient } from "../src";
+
+export const bucket = new BucketClient({
+  secretKey: process.env.BUCKET_SECRET_KEY || "",
```
pavkam, line 4: We need to settle on the fallback flags and maybe add some here for reference?

packages/node-sdk/example/app.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const todos = ["Buy milk", "Walk the dog"];
+
+app.get("/", (req, res) => {
+  res.locals.bucketUser.track("Front Page Viewed");
+  res.send("Hello World");
+});
+
+app.get("/todos", (req, res) => {
+  const showTodosFlag = await res.locals.bucketUser.getFlag("show-todos");
```
pavkam, line 32: Wait, did I add a separate `getFlag`? I though not. Also, we discussed the `global` flag type, I opted to make the `getFlags` generic. Should we exemplify that by declaring a `type AppFlags = 'show-todos' | 'something'` and

Should we make the `BucketClient` generic or just the flag getting function?

packages/node-sdk/example/bucket.ts
roncohen: with modern Node.js it should work.

packages/node-sdk/example/bucket.ts
roncohen: yes, good idea

roncohen:
As usual, fighting the "build system" to get everything compiled etc.

PR #153, opened 2024-07-23
Browser sdk

roncohen:
Introducing the Browser SDK. I will incorporate https://github.com/bucketco/bucket-javascript-sdk/pull/144 separately.

- Switched to vite for building instead of webpack for both React SDK and Browser SDK
- Updated React SDK to use the new Browser SDK
- Most code is lifted from the tracking SDK. 
- Replaced Nock with MSW which is more maintained and browser focused.


packages/browser-sdk/src/client.ts
```diff
@@ -0,0 +1,260 @@
...
+  TrackedEvent,
+  User,
+} from "./feedback/feedback";
+import { FlagsOptions, FlagsClient } from "./flags/flags";
+import { Logger, loggerWithPrefix, quietConsoleLogger } from "./logger";
+import { HttpClient } from "./httpClient";
+import * as feedbackLib from "./feedback/ui";
+import { BucketContext } from "./context";
+
+const isMobile = typeof window !== "undefined" && window.innerWidth < 768;
```
pavkam, line 18: Is this the best way to detect that?

packages/browser-sdk/src/feedback/feedback.ts
```diff
@@ -0,0 +1,456 @@
...
+
+    /**
+     * Add your own custom translations for the feedback form.
+     * Undefined translation keys fall back to english defaults.
+     */
+    translations?: Partial<FeedbackTranslations>;
+  };
+};
+
+export type User = {
```
pavkam, line 39: not sure these types should be here?

packages/browser-sdk/src/feedback/feedback.ts
```diff
@@ -0,0 +1,456 @@
...
+  //   feedbackId,
+  //   userId: String(this.context.user.id),
+  //   featureId,
+  //   score,
+  //   companyId,
+  //   comment,
+  //   promptId,
+  //   question,
+  //   promptedQuestion,
+  // TODO: check that source is always set
```
pavkam, line 237: just a note to remind to either remove code or fix the TODO

packages/browser-sdk/src/feedback/feedback.ts
```diff
@@ -0,0 +1,456 @@
...
+        sseHost: this.sseHost,
+      });
+
+      this.logger.debug(`feedback prompting connection established`);
+    } catch (e) {
+      this.logger.error(`error initializing feedback prompting`, e);
+      return;
+    }
+    return;
+    // return channel;
```
pavkam, line 311: reminder to figure out this return

packages/browser-sdk/src/feedback/feedback.ts
```diff
@@ -0,0 +1,456 @@
...
+  }
+
+  async triggerFeedbackPrompt(
+    userId: string,
+    message: FeedbackPrompt,
+    completionHandler: FeedbackPromptCompletionHandler,
+  ) {
+    let feedbackId: string | undefined = undefined;
+
+    // TODO: still needed?
```
pavkam, line 353: is needed if we allow switching users after the SDK has been initialized. Otherwise, not really needed.

packages/browser-sdk/src/client.ts
roncohen: Probably not, but it's from the tracking-sdk so keeping it as-is for now.

packages/browser-sdk/src/feedback/feedback.ts
roncohen: great point! moved into `client.ts`

pavkam (approved):
Let's get this in.

PR #156, opened 2024-07-30
feat: recursive filter evaluation

matus-vacula:


roncohen:
Is there a reason PercentageRolloutFilter is now a filter? It's still a property on the `rule` right?

matus-vacula:
Since there can be a feature flag filter anywhere in the filter tree we have to also have the rollout filter in the matching place when evaluating the boolean algebra.

roncohen:
right, makes sense.

PR #163, opened 2024-08-01
Fix flakiness in tests

roncohen:
Was only failing on CI. Increasing the timeout seems to have worked, although it's not ideal


roncohen:
OK this is ready for a review. Finally tracked down the bugger. Obvious in hindsight.

pavkam (approved):

PR #171, opened 2024-08-09
feat: track function for `useFeature`

roncohen:
Introduce a `track` function return from `useFeature` to let you easily track feature usage.

### DIFF
--- a/packages/react-sdk/src/index.tsx
+++ b/packages/react-sdk/src/index.tsx
@@ -206,28 +206,39 @@ export function BucketProvider({
     requestFeedback,
   };
 
-  if (featuresLoading && loadingComponent) {
-    return loadingComponent;
-  }
-
-  return <ProviderContext.Provider children={children} value={context} />;
+  return (
+    <ProviderContext.Provider
+      children={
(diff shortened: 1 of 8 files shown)


packages/react-sdk/src/index.tsx
```diff
@@ -206,28 +206,39 @@ export function BucketProvider({
...
-    useContext<ProviderContextType>(ProviderContext).features;
+  const {
+    features: { features, isLoading },
+    track,
+  } = useContext<ProviderContextType>(ProviderContext);
 
   const isEnabled = features[key] ?? false;
 
-  return { isLoading: isLoading, isEnabled };
+  return { isLoading: isLoading, isEnabled, track: () => track(key) };
```
matus-vacula, line 241: Just thinking out loud, don't we want to provide an optional argument for event attributes?

matus-vacula (approved):

packages/react-sdk/src/index.tsx
roncohen: This `track` function is for the plug-play scenario. My thinking was to minimize the API for now and then add if we ever find a use for it.

PR #176, opened 2024-08-13
Simplify fetching logic

roncohen:
- Remove retry-logic that was too complicated. We will no longer cache negative responses at all.
- Increase expire time from 7 to 30 days.
- Don't block on LiveSatisfaction initialization

### DIFF
--- a/packages/browser-sdk/src/feature/features.ts
+++ b/packages/browser-sdk/src/feature/features.ts
@@ -144,6 +144,10 @@ export class FeaturesClient {
 
   async initialize() {
     const features = (await this.maybeFetchFeatures()) || {};
+    this.setFeatures(features);
+  }
+
+  private setFeatures(features: APIFeaturesResponse) {
     const proxiedFeatures = maskedProxy(features, (fs, key) => {
       this.sendCheckEvent({
         key,
@@ -190,6 +194,7 @@ export class FeaturesClient {
             this.cache.set(cacheKey, {
               features,
             });
+            this.setFeatures(features);
           })
           .catch(() => {
             // we don't care about the result, we just want to re-fetch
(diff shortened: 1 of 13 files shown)


packages/browser-sdk/src/feature/features.ts
```diff
@@ -177,40 +178,60 @@ export class FeaturesClient {
...
+          .catch(() => {
+            // we don't care about the result, we just want to re-fetch
+          });
         return cachedItem.features;
       }
+    }
+
+    // if there's no cached item or there is a stale one but `staleWhileRevalidate` is disabled
+    // try fetching a new one
+    const fetchedFlags = await this.fetchFeatures();
```
pavkam, line 208: I see a mixture of "flags" and "features" in the names. Probably better to switch to features in every name and comment for clarity

pavkam (approved):
Just to confirm: we have opted out from having an immediate retry mechanism. Instead we will poll for flags in background but also try to get fresh ones anytime when we can without having a "max tries"

packages/browser-sdk/src/feature/features.ts
roncohen: good catch. Missed that one 👍

roncohen:
> Just to confirm: we have opted out from having an immediate retry mechanism. Instead we will poll for flags in background but also try to get fresh ones anytime when we can without having a "max tries"

For feature/flag fetching, there was never an immediate retry actually. It would only try on page loads. If there was multiple failed loads within 60s it would eventually give up and cache a "negative" result and would then use that for 60s. This was super confusing and the code was error prone. There's also no background refresh yet. I figured we'd go straight to #165 instead.

PR #177, opened 2024-08-14
React + Browser v1.1.0

roncohen:
After https://github.com/bucketco/bucket-javascript-sdk/pull/176 goes in, we'll cut this new version.

### DIFF
--- a/packages/browser-sdk/package.json
+++ b/packages/browser-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/browser-sdk",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "packageManager": "yarn@4.1.1",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


packages/browser-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/browser-sdk",
-  "version": "1.0.1",
+  "version": "1.1.0",
```
pavkam, line 3: are thy ready for prime time? Should we prefix them all with `alpha.x` instead?

pavkam (approved):

packages/browser-sdk/package.json
roncohen: they are ready. Also, it's too late, already shipped 1.0 ;)

PR #183, opened 2024-08-23
[NodeSDK] Simplify the API

roncohen:
Moved to an API with two classes: `BucketClient` and `BoundBucketClient` which is bound to a user/company for ease of use.

### DIFF
--- a/packages/node-sdk/src/client.ts
+++ b/packages/node-sdk/src/client.ts
@@ -30,7 +30,6 @@ import {
   maskedProxy,
   ok,
 } from "./utils";
-import { isArray } from "util";
 
 /**
  * The SDK client.
@@ -47,14 +46,6 @@ export class BucketClient {
     fallbackFeatures?: Record<keyof TypedFeatures, InternalFeature>;
     featuresCache?: Cache<FeaturesAPIResponse>;
   };
-
-  /**
-   * Creates a new SDK client.
-   *
-   * @param options - The options for the client.
-   * @throws An error if the options are invalid.
-   **/
-  constructor(options: ClientOptions);
 
   /**
    * Creates a new SDK client.
@@ -471,7 +462,7 @@ export class BucketClient {
 
       // TODO: use the bulk endpoint
       evaluated.forEach(async (res) => {
-        await this.sendFeatureEvent({
+        this.sendFeatureEvent({
           action: "evaluate",
           key: res.feature.key,
           targetingVersion: keyToVersionMap.get(res.feature.key),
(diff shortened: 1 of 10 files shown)


roncohen:
WDYT @pavkam ?

packages/node-sdk/src/client.ts
```diff
@@ -546,41 +445,35 @@ export class BucketClientClass {
...
 
       const evaluated = featureDefinitions.features.map((feature) =>
-        evaluateTargeting({ context: mergedContext, feature }),
+        evaluateTargeting({ context, feature }),
       );
 
+      // TODO: use the bulk endpoint
       evaluated.forEach(async (res) => {
-        await this.sendFeatureEvent({
+        this.sendFeatureEvent({
```
pavkam, line 464: the lambda is still marked as `async`, but not needed

packages/node-sdk/src/client.ts
```diff
@@ -611,9 +504,140 @@ export class BucketClientClass {
...
+        this._config.logger?.error(
+          `failed to send check event for "${key}": ${err}`,
+          err,
+        );
       });
 
-      return features[key].isEnabled;
+      const feature = features[key];
+
+      return {
```
pavkam, line 516: is the `maskedProxy` smart to ignore the `track()` property?

packages/node-sdk/src/client.ts
```diff
@@ -546,41 +445,35 @@ export class BucketClientClass {
...
-        ...this._user.attrs,
-      },
-      company: this._company && {
-        id: this._company.companyId,
-        ...this._company.attrs,
-      },
-      other: this._otherContext,
-    };
-
+  public getFeatures(context: Context): TypedFeatures {
```
pavkam, line 448: how does that work? You chave to do `boundClient.getFeatures(boundClient.context())`?

pavkam: I a missing some code I think as I do not see the connection between the `BoundClient` and the unbound one in terms of passing the context to these features.

roncohen: `BoundBucketClient` has a `context` and a `BucketClient` and when you call anything on the `BoundBucketClient` it'll just take it's context and call the corresponding method on the `BucketClient` passing along the context.

packages/node-sdk/src/client.ts
roncohen: no. It'll send the "check" event even if you call `track`. I think it's OK for now because it stil indicates that there's a reference in the code.

packages/node-sdk/src/client.ts
pavkam: I haven't noticed the:
```
  public getFeatures() {
    return this._client.getFeatures(this._context);
  }
```

before, hm

pavkam (approved):

PR #188, opened 2024-08-25
[Browser SDK] send user/company on init

roncohen:
The BrowserSDK needs to send user/company on init. Moving it out of the React SDK and into the Browser SDK.

### DIFF
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -149,13 +149,13 @@ export class BucketClient {
     await this.featuresClient.initialize();
 
     if (this.context.user) {
-      this.user().catch((e) => {
+      this.user(this.context.user).catch((e) => {
         this.logger.error("error sending user", e);
       });
     }
 
     if (this.context.company) {
-      this.company().catch((e) => {
+      this.company(this.context.company).catch((e) => {
         this.logger.error("error sending company", e);
       });
     }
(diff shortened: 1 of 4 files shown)


packages/browser-sdk/src/client.ts
```diff
@@ -148,6 +148,18 @@ export class BucketClient {
 
     await this.featuresClient.initialize();
 
+    if (this.context.user) {
+      this.user(this.context.user).catch((e) => {
```
pavkam, line 152: I wonder why my lsp complains about the unawaited promises and yours doesnt. I usually add the special void keyword in front to make typescript understand I don't want to await.

roncohen: agreed, pretty weird. I've learned from your comments to always add the `void`. These two are awaited though :)

pavkam: The catch call returns another promise though

roncohen: `catch` is enough AFAICT? : https://typescript-eslint.io/rules/no-floating-promises/

pavkam: Interesting. I think I've configured my TS LSP to be stricter and that's why it won't let me be in such cases. But i agree, it's enough.

pavkam (approved):

PR #196, opened 2024-09-02
fix: do not send 'undefined' user/company/other

roncohen:
Using the Browser SDK directly (instead of through something like the React SDK) we'd sometimes send a `other: undefined` context along with the normal context if no "other" context was given.  I don't think it was a problem, but it isn't nice :)


packages/browser-sdk/src/feature/features.ts
```diff
@@ -92,6 +92,8 @@ export function flattenJSON(obj: Record<string, any>): Record<string, any> {
       for (const flatKey in flat) {
         result[`${key}.${flatKey}`] = flat[flatKey];
       }
+    } else if (typeof obj[key] === "undefined") {
```
pavkam, line 95: JSON has no concept of `undefined` so it would not be serialized anyway, but as you said - let's play nice :)

pavkam (approved):

packages/browser-sdk/src/feature/features.ts
roncohen: We don’t actually serialize to JSON in this case. The context of flattened and then sent as search parameters.

PR #202, opened 2024-09-10
feat(openfeature-node-provider): initial commit

roncohen:
Introduce the Bucket Node.js provider for OpenFeature.


packages/openfeature-node-provider/README.md
```diff
@@ -0,0 +1,114 @@
+# Bucket Node.js Provider
```
pavkam, line 1: Maybe mention it's Open feature provider

packages/openfeature-node-provider/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+This provider is an OpenFeature implementation for [Bucket](https://bucket.co) feature management service.
```
pavkam, line 3: Adapter rather

packages/openfeature-node-provider/README.md
```diff
@@ -0,0 +1,114 @@
...
+const enterpriseFeatureEnabled = await client.getBooleanValue(
+  "enterpriseFeature",
+  false,
+  requestContext,
+);
+```
+
+## Translating Evaluation Context
+
+Bucket uses a context object in the following shape:
```
pavkam, line 58: Of the following shape

packages/openfeature-node-provider/README.md
```diff
@@ -0,0 +1,114 @@
...
+  company?: { id: string; [k: string]: any };
+  /**
+   * The other context. This is used for any additional context that is not related to user or company.
+   */
+  other?: Record<string, any>;
+};
+```
+
+To use the Bucket Node.js OpenFeature provider, you must convert your OpenFeature contexts to Bucket contexts.
+You do that by supplying a context translation function which takes the Open Feature context and returns
```
pavkam, line 82: You can achieve this

packages/openfeature-node-provider/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+      }
+    : undefined;
+
+  return {
+    user,
+    company,
+  };
+};
+
+export class BucketNodeProvider implements Provider {
```
pavkam, line 47: Should we add some jsdocs?

packages/openfeature-node-provider/README.md
roncohen: what do you mean exactly?

packages/openfeature-node-provider/src/index.ts
roncohen: We could, but you'll realistically be operating against the OpenFeature facade/interface rather than this client directly, so I don't think it's super important.

packages/openfeature-node-provider/README.md
pavkam: Sorry, was on the phone when writing so was short on details, I meant, in order to disambiguate with the actual "provider", we should call this library an "Adapter for Open Feature" instead.

roncohen: In OpenFeature nomenclature, this is called a "provider": https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers

pavkam: Ah, my bad

pavkam (approved):
🚀

PR #213, opened 2024-09-16
Node.js SDK batching support

pavkam:
This PR adds a new class into the mix. This class is used as a "batch buffer" that will store all given data and then flush it at a given time interval or when the buffer is full.

There is also a secondary retry buffer that is being flushed on a different cadence and items in there are dropped out after a number of tries.

The `BucketClient` has been updated to route `user`, `company` and `feature-flag-event` payloads into the buffer instead of calling the individual API endpoints.


packages/node-sdk/src/utils.ts
```diff
@@ -112,3 +114,123 @@ export function decorateLogger(prefix: string, logger: Logger): Logger {
     },
   };
 }
+
+export class AccumulatingBuffer<T> {
```
roncohen, line 118: don't need this, right?

packages/node-sdk/src/batch-buffer.ts
```diff
@@ -0,0 +1,176 @@
...
+  private logger?: Logger;
+  private maxSize: number;
+  private intervalMs: number;
+  private retryIntervalMs: number;
+  private maxRetries: number;
+  private timer: NodeJS.Timeout | null = null;
+  private retryTimer: NodeJS.Timeout | null = null;
+
+  /**
+   * Creates a new `AccumulatingBuffer` instance.
```
roncohen, line 27:
```suggestion
   * Creates a new `BatchBuffer` instance.
```

pavkam: Thanks

packages/node-sdk/src/utils.ts
pavkam: Yeah, just left it there by mistake

pavkam:
* remove retries from batching,
* add companyId in bound client to track,
* track should be buffered as well.

roncohen (commented):
generally looks good. As discussed, let's pull the retry stuff out for now and handle regular track events like other events. If you need instant auto. feedback you'll have to call `flush` manually.

roncohen (approved):
Nice!

One more thing needed to get `companyId` into track calls automatically. 

This [line](https://github.com/bucketco/bucket-javascript-sdk/pull/213/files/75c45b24516e8c710948a44f5289a5438eb56911..8ad34693ba3e7bd7783a04d8d93c9e325ccdcad5#diff-d3e82fa914c46cab66ff3b2d56ba05e3b91b1653bee16a198a665488ca50e271R607) needs to change into something like

```ts
await this.track(userId, key, { companyId: context.company?.id });
```

packages/node-sdk/README.md
```diff
@@ -76,8 +80,8 @@ messages could be waiting to be sent, and thus, would be discarded if the buffer
 A naive example:
 
 ```ts
-["SIGINT", "SIGTERM", "SIGQUIT"].forEach((signal) => {
-  process.on(signal, () => client.flush());
+process.on("beforeExit", (code) => {
```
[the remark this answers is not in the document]
roncohen: I just tried this and unfortunately `process.on('beforeExit',...` doesn't actually work. Looks like we need something close to what you had before:
```ts
process.on("SIGINT", () => {
  console.log("Flushing batch buffer...");
  client.flush().then(() => "flushed");
  process.exit(0)
});
```

apologies for the noise here.

pavkam:
I did that on the bound client only. You reckon we should implicitly do that on unbound one as well?

packages/node-sdk/README.md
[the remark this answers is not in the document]
pavkam: No worries. I'll update. And thanks for the tip on exit. Didn't know we completely override the signal handler and stop the process from exiting.

pavkam:
> > I did that on the bound client only. You reckon we should implicitly do that on unbound one as well?
> 
> 
> 
> I think it makes sense for the `track` function that is returned as a result of calling getFeatures with a specific context.

I see, I forgot about that one actually. But yes, makes sense. I'll add it.

pavkam:
> Nice!
> 
> One more thing needed to get `companyId` into track calls automatically.
> 
> This [line](https://github.com/bucketco/bucket-javascript-sdk/pull/213/files/75c45b24516e8c710948a44f5289a5438eb56911..8ad34693ba3e7bd7783a04d8d93c9e325ccdcad5#diff-d3e82fa914c46cab66ff3b2d56ba05e3b91b1653bee16a198a665488ca50e271R607) needs to change into something like
> 
> ```ts
> await this.track(userId, key, { companyId: context.company?.id });
> ```

Done-zo

PR #216, opened 2024-09-18
chore(NodeSDK): 1.1.0

pavkam:
Adding a minor bump.


roncohen (approved):
Lets make sure the commit message ends up being something like

chore(NodeSDK): 1.1.0

PR #220, opened 2024-10-02
feat(node-sdk): introduce `getFeature`

roncohen:
Introduce a `getFeature` method which should be the preferred way of getting a specific feature.
I also fixed a few things I found along the way:
- only send `check` event when `isEnabled` is get'et
- add `flush()` to BoundBucketClient for convenience
- docs improvements

### DIFF
--- a/packages/node-sdk/README.md
+++ b/packages/node-sdk/README.md
@@ -60,16 +60,16 @@ const boundClient = client.bindClient({
 });
 
 // get the current features (uses company, user and custom context to evaluate the features).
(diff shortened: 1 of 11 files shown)


pavkam:
Why not add an optional argument to the original function?

roncohen:
I reverted the changes from `getFeature` -> `getFeaturesRaw` so there are no longer any backwards incompatible changes.

roncohen:
thanks @Swiftwork !

PR #226, opened 2024-10-28
feat: `enableTracking` makes it possible to disable tracking

roncohen:
In situations where you do not want to send events or track user/company details on behalf of a user but still want to use them for feature targeting, you should be able to disable it. This is useful for example when and admin is impersonating another user.

### DIFF
--- a/packages/react-sdk/README.md
+++ b/packages/react-sdk/README.md
@@ -80,7 +80,7 @@ import { BucketProvider } from "@bucketco/react-sdk";
   <BucketProvider>
   ```
 
-- `impersonating` (default: `false`): Set to `true` to stop sending tracking events and user/company updates to Bucket. Also stops automated feedback surveys.
+- `enableTracking` (default: `true`): Set to `false` to stop sending tracking events and user/company updates to Bucket.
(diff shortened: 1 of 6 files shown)


pavkam (approved):
Look good

Swiftwork (approved):

packages/react-sdk/README.md
```diff
@@ -80,6 +80,8 @@ import { BucketProvider } from "@bucketco/react-sdk";
   <BucketProvider>
   ```
 
+- `impersonating` (default: `false`): Disables sending tracking events and user/company updates to Bucket and disabled automated feedback surveys. Set to `true` when impersonating another user.
```
matus-vacula, line 83: The wording should be "enables".

roncohen: OK i can see how this is confusing. If you set it to something other than the default then it'll disable sending tracking events .... Needs a rewording.

roncohen: OK let me know if this is better.

pavkam (approved):

PR #227, opened 2024-10-29
feat: getFeaturesRemote

matus-vacula:


packages/node-sdk/src/client.ts
```diff
@@ -647,6 +647,45 @@ export class BucketClient {
...
+  ): Promise<TypedFeatures> {
+    const context = additionalContext || {};
+    if (userId) {
+      context.user = { id: userId };
+    }
+    if (companyId) {
+      context.company = { id: companyId };
+    }
+    const params = new URLSearchParams(flattenJSON({ context })).toString();
+    return this.get<EvaluatedFeaturesAPIResponse>(
```
pavkam, line 671: can we use imperative async code to be in line with the rest of the code.

packages/node-sdk/src/client.ts
```diff
@@ -647,6 +647,45 @@ export class BucketClient {
...
   }
+
+  /**
+   * Gets evaluated features with the usage of remote context.
+   * This method triggers a network request every time it's called.
+   *
+   * @param additionalContext
+   * @returns evaluated features
+   */
+  public async getFeaturesRemote(
```
pavkam, line 658: wonder if we can make `getFeatures` work in both cases based on a boolean. Typescript should allow us to change the return value to either promise or not based on the types of args. Or maybe that's too confusing

matus-vacula: That probably won't work because one is async and the other one not.

pavkam (approved):

roncohen (approved):

PR #232, opened 2024-11-05
fix: set staletime to zero to avoid confusion

roncohen:
At initialization time (page load) we fetch features. There's a cache of 60s today which means if you loaded the page previously within 60s we will not refetch. This can be cause for confusion. After changing targeting rules, people expect that after a hard refresh features will have updated. If they wait a bit and then hard refresh they do change which makes them end up looking random. We added this cache to avoid help in the case of non-SPA which would fetch features with every load and incur the overhead of the fetch request every time. However, since the norm is an SPA I'm changing it to default to zero. Folks with an non-SPA can still set the stale time to 60s and avoid the refetch with every load.

### DIFF
--- a/packages/browser-sdk/README.md
+++ b/packages/browser-sdk/README.md
@@ -72,7 +72,7 @@ See [example/browser.html](example/browser.html) for a working example:
 
 ### Init options
 
-Supply these to the constructor call (3rd argument)
+Supply these to the constructor call:
 
 ```ts
 {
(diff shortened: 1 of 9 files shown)


packages/browser-sdk/README.md
```diff
@@ -85,6 +85,8 @@ Supply these to the constructor call (3rd argument)
     fallbackFeatures?: string[]; // Enable these features if unable to contact bucket.co
     timeoutMs?: number; // Timeout for fetching features
     staleWhileRevalidate?: boolean; // Revalidate in the background when cached features turn stale to avoid latency in the UI
+    staleTimeMs?: // at initialization time features are loaded from the cache unless they have gone stale. Defaults to 0 which means the cache is disabled. Increase in the case of a non-SPA.
```
pavkam, line 88: should we make it configurable here but keep it at 60 and instead set these to zero in react SDKs?

roncohen: great point. I went back and forth with myself on it. I think for the sake of clarity we should just set it to zero in the BrowserSDK too because people will be building SPAs with that as the basis for anything that isn't React.

pavkam (approved):

PR #233, opened 2024-11-06
Node sdk needs to flush the rate limiter to avoid memory

pavkam:


packages/node-sdk/src/rate-limiter.ts
```diff
@@ -0,0 +1,82 @@
...
+      eventsByKey = {};
+    } else {
+      const expiredAfter = Date.now() - windowSizeMs;
+
+      for (const key in eventsByKey) {
+        const events = eventsByKey[key];
+        const last = events.length ? events[events.length - 1] : undefined;
+
+        if (last && last < expiredAfter) {
+          delete eventsByKey[key];
```
roncohen, line 47: this only works for `eventLimit=1` right? Since we don't need anything else than `eventLimit=1` lets just simplify and assume that `eventLimit=1`. We can remove the lists inside `eventsByKey`?

pavkam: Actually it should have worked with many case. But you are correct, why bother. Just the size of the window is enough.

roncohen (approved):
LGTM 👍

packages/node-sdk/src/rate-limiter.ts
```diff
@@ -0,0 +1,63 @@
...
+        if (lastAllowedTimestamp < expiredAfter) {
+          delete lastAllowedTimestampsByKey[key];
+        }
+      }
+    }
+  }
+
+  function isAllowed(key: string): boolean {
+    clearIntervalId =
+      clearIntervalId || setInterval(() => clear(false), windowSizeMs);
```
roncohen, line 46: I think could be `setTimeout` since we're not using the intercal functionality?

packages/node-sdk/src/rate-limiter.ts
```diff
@@ -0,0 +1,63 @@
...
+  function clear(all: boolean): void {
+    if (clearIntervalId) {
+      clearInterval(clearIntervalId);
+      clearIntervalId = undefined;
+    }
+
+    if (all) {
+      lastAllowedTimestampsByKey = {};
+    } else {
+      const expiredAfter = Date.now() - windowSizeMs;
```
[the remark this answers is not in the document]
pavkam: :)

packages/node-sdk/src/rate-limiter.ts
pavkam: Actually wee need that, but the clearing was too aggressive, should only kill the timeout if there's no smoke keys left in the cache. I rearranged the code.

If we keep setting a new interval each time `isAllowed` is invoked then we risk into never actually running the timer since it would be renewed every time a check is done (which could be less than once a minute).

We could also invoke `clear(false)` in the `isAllowed` and only start a timer if needed, but potato/potatoh :)

roncohen (approved):
LGTM

PR #235, opened 2024-11-07
feat(node-sdk): Implement automatic attribute sending and tracking toggle

pavkam:
This PR adds automatic sending of `user` and `company` events to Bucket when either `bindClient` or `getFeature(s)` calls are made.

Also introduces `enableTracking` in `Context` to allow disabling to send `event`, `user` and `company` when the user is not interested in sending them (e.g. impersonating other users).

Employing the rate-limiter on `user` and `company` calls to limit the amount of potential calls made to `Bucket` servers.


packages/node-sdk/src/client.ts
```diff
@@ -956,13 +1046,14 @@ export class BoundBucketClient {
    * @param context User/company/other context to bind to the client object
    * @returns new client bound with the additional context
    */
-  public bindClient({ user, company, other }: Context) {
+  public bindClient({ user, company, other, enableTracking }: Context) {
```
roncohen, line 1049: Context has a very specific meaning: it's what you run the targeting rules on. `enableTracking` is just configuration for this new Client. Let's keep 'em separate. In other words, Context stays the way it is but `bindClient` and `getFeature`/`getFeatures` take an object `Context & {enableTracking: boolean}`

packages/node-sdk/src/client.ts
```diff
@@ -956,13 +1046,14 @@ export class BoundBucketClient {
...
     // merge new context into existing
     const newContext = {
```
roncohen, line 1051:
```suggestion
    const boundConfig = {
```

packages/node-sdk/src/client.ts
```diff
@@ -583,6 +626,12 @@ export class BucketClient {
   }
 
   private _getFeatures(context: Context): Record<string, RawFeature> {
```
roncohen, line 628: something like this would simplify:
```suggestion
  private _getFeatures({enableTracking=true, ...context}: Context & {enableTracking: boolean}): Record<string, RawFeature> {
```

packages/node-sdk/src/client.ts
```diff
@@ -583,6 +626,12 @@ export class BucketClient {
...
+    void this.syncContext(context);
```
roncohen, line 629:
```suggestion
if (enableTracking=true) void this.syncContext(context)
```

roncohen (commented):
added some refactoring suggestions

packages/node-sdk/src/client.ts
```diff
@@ -367,6 +367,67 @@ export class BucketClient {
...
+    const promises: Promise<void>[] = [];
+    if (
+      context.company &&
+      this._config.rateLimiter.isAllowed(
+        JSON.stringify(context.company, null, 0),
+      )
+    ) {
+      const { id: _, ...attributes } = context.company;
+      promises.push(
+        this.updateCompany(context.company.id, {
```
roncohen, line 401: lets check `isAllowed` inside `updateCompany`, `updateUser` etc. instead. Same for `enableTracking`. Should simplify things

roncohen:
> currently, the evaluate and check events for flags are not suppressed

What do you mean? `sendFeatureEvent` checks `isAllowed`.

pavkam:
> > currently, the evaluate and check events for flags are not suppressed
> 
> What do you mean? `sendFeatureEvent` checks `isAllowed`.

I mean are they covered by `enableTracking`?

packages/node-sdk/src/client.ts
pavkam: Good one

packages/node-sdk/src/client.ts
pavkam: I would prefer to keep it as is due to a log inside, otherwise I'd have to copy-paste the same one in other parts.

roncohen:
There are still tons of references to "context" which include `enableTracking`. Context is just {user, company, other} so we need to clean everything else up.

Screenshot 2024-11-09 at 14 11 02

pavkam:
> There are still tons of references to "context" which include `enableTracking`. Context is just {user, company, other} so we need to clean everything else up.
> 
> 
> 
> Screenshot 2024-11-09 at 14 11 02
> 
> 

That is a ContextWithTracking which is a local type to avoid repeating the same union in all the code. Should we rename the args? Or what do you have in mind?

roncohen:
> Should we rename the args? Or what do you have in mind?

yes, `enableTracking` has nothing to do with `context`:
- `BoundBucketClient._context` member variable contains a `enableTracking`
- validation function `checkContext` also takes `enableTracking`
- and so on

additionally, there's no point in making `enableTracking` optional. Let's make it default to `true` so we can remove a bunch of checks.

roncohen:
> I mean are they covered by `enableTracking`?

Good point. Since `enableTracking=false` was introduced to allow for impersonation, let's also ignore flag events when enableTracking is off. I'll update the BrowserSDK to do the same

pavkam:
> > Should we rename the args? Or what do you have in mind?
> 
> 
> 
> yes, `enableTracking` has nothing to do with `context`:
> 
> - `BoundBucketClient._context` member variable contains a `enableTracking`
> 
> - validation function `checkContext` also takes `enableTracking`
> 
> - and so on
> 
> 
> 
> additionally, there's no point in making `enableTracking` `undefined | boolean`. Let's make it default to `true` so we can remove a bunch of checks.

should we then, extract it as a secondary arg to those methods? Otherwise i don't see the difference keeping it in the _context. It's a private anyway and we'd still need to reconstruct the context plus this flag when calling the real client methods.

A user of our SDK won't see the difference if it's explicitly specified as an union object or if it's a local type in the file.

The only real difference is if we make it as an optional argument to those methods that can use it.

I'm a bit lost here in what you are seeing that I'm not.

pavkam:
> > I mean are they covered by `enableTracking`?
> 
> Good point. Since `enableTracking=false` was introduced to allow for impersonation, let's also ignore flag events when enableTracking is off. I'll update the BrowserSDK to do the same

Updated the code as discussed. Have a look.

packages/node-sdk/src/client.ts
```diff
@@ -77,6 +78,17 @@ type BulkEvent =
...
 
+/**
+ * A context with tracking option.
+ **/
+type ContextWithTracking = Context & {
+  /**
+   * Enable tracking for the context.
+   * If set to `false`, tracking will be disabled for the context. Default is `true`.
+   */
+  enableTracking?: boolean;
```
roncohen, line 89: Looks like we can remove the question mark here since we set the default everywhere.

packages/node-sdk/README.md
```diff
@@ -47,46 +47,84 @@ client.initialize().then({
...
 // warning: if the `huddle` feature does not exist because it wasn't created in Bucket
-// or because the client was unable to reach our servers for some reason, this will cause an exception:
+// or because the client was unable to reach our servers for some reason, this will
+// cause an exception:
 const { isEnabled } = boundClient.getFeatures()["huddle"];
 ```
 
+## Opting out of tracking
+
+There are use cases in which one does not want to be sending `user`, `company` and
```
roncohen, line 98: Rest of the README is using `you` instead of `one`.

packages/node-sdk/README.md
```diff
@@ -47,46 +47,84 @@ client.initialize().then({
...
+// not send `track` events either.
+const { isEnabled, track } = boundClient.getFeature("user-menu");
+if (isEnabled) {
+  track();
+}
+```
+
+Another way way to disable tracking without employing a bound client is to call `getFeature()`
+or `getFeatures()` by supplying `enableTracking: false` in the context passed to
+these functions.
```
roncohen, line 121:
```suggestion
Another way way to disable tracking without employing a bound client is to call `getFeature()`
or `getFeatures()` by supplying `enableTracking: false` in the arguments passed to
these functions.
```

packages/node-sdk/src/client.ts
```diff
@@ -371,6 +389,57 @@ export class BucketClient {
...
+   */
+  private async syncContext({
+    enableTracking = true,
+    ...context
+  }: ContextWithTracking) {
+    checkContextWithTracking({ enableTracking, ...context });
+
+    if (!enableTracking) {
+      this._config.logger?.debug(
+        "feature.syncContext(): tracking disabled from context, not updating user/company",
```
roncohen, line 406:
```suggestion
        "tracking disabled, not updating user/company",
```

packages/node-sdk/src/client.ts
```diff
@@ -701,26 +782,36 @@ export class BucketClient {
...
-        await this.track(userId, key, {
-          companyId: context.company?.id,
-        });
+        if (enableTracking) {
+          await this.track(userId, key, {
+            companyId: context.company?.id,
+          });
+        } else {
+          this._config.logger?.debug(
+            "feature.track(): tracking disabled from context, not tracking event",
```
roncohen, line 791:
```suggestion
            "tracking disabled, not tracking event",
```

packages/node-sdk/src/client.ts
```diff
@@ -953,19 +1039,26 @@ export class BoundBucketClient {
...
 
-    const userId = this._context.user?.id;
+    const userId = this._options.user?.id;
 
     if (!userId) {
       this._client.logger?.warn("no user set, cannot track event");
       return;
     }
 
+    if (this._options.enableTracking === false) {
```
roncohen, line 1049: if `enableTracking` becomes mandatory, this can be simplified.

roncohen (approved):
minor remarks. Thank you for your patience in this!

packages/node-sdk/src/client.ts
```diff
@@ -976,16 +1069,22 @@ export class BoundBucketClient {
    * @param context User/company/other context to bind to the client object
    * @returns new client bound with the additional context
    */
-  public bindClient({ user, company, other }: Context) {
+  public bindClient({
+    user,
+    company,
+    other,
+    enableTracking = true,
```
roncohen, line 1076: I think in this case where you're creating `BucketBoundClient` from an existing `BucketBoundClient` we should probably default to inheriting the setting that you have for `enableTracking` and not default to true. The situation i have in mind is that you want to be able to set a basic `BucketBoundClient` with `enableTracking=false` and then at a later point in the middleware series just use whatever setting you had for `enableTracking` further up in the chain.

In short, let's make `enableTracking` optional here like you had previously. Then line 1084 below also makes more sense: use what was given, otherwise use the "parent" `BucketBoundClient`'s value.

packages/node-sdk/src/client.ts
pavkam: You're right, I had to use `interface` instead of `type`

packages/node-sdk/README.md
pavkam: Switched to `you`

packages/node-sdk/README.md
pavkam: Danke

packages/node-sdk/README.md
```diff
@@ -47,46 +47,84 @@ client.initialize().then({
...
 
 ```ts
 // warning: if the `huddle` feature does not exist because it wasn't created in Bucket
-// or because the client was unable to reach our servers for some reason, this will cause an exception:
+// or because the client was unable to reach our servers for some reason, this will
+// cause an exception:
 const { isEnabled } = boundClient.getFeatures()["huddle"];
 ```
 
+## Opting out of tracking
```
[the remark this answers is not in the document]
pavkam: done

packages/node-sdk/src/client.ts
```diff
@@ -609,85 +675,100 @@ export class BucketClient {
...
         return this._config.fallbackFeatures || {};
       }
+
       featureDefinitions = fetchedFeatures.features;
     }
 
-    const keyToVersionMap = new Map<string, number>(
-      featureDefinitions.map((f) => [f.key, f.targeting.version]),
-    );
+    let evaluatedFeatures: Record<keyof TypedFeatures, RawFeature> =
```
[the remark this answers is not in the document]
pavkam: oh yea, you're right, merge issue

packages/node-sdk/src/client.ts
pavkam: true

PR #238, opened 2024-11-08
Various docs improvements

roncohen:
- Removed reference to fallback flags. Fallback flags was a distraction for folks and it's not really feasible to maintain that list.
- Explained how to get typesafe feature names
- Show how to use with Express

### DIFF
--- a/packages/node-sdk/README.md
+++ b/packages/node-sdk/README.md
@@ -162,7 +162,7 @@ import bucket from "./bucket";
 import express from "express";
 import { BoundBucketClient } from "@bucketco/node-sdk";
 
-// Augment the Express types to include the `boundBucketClient` property on the `res.locals` object
+// Augment the Express types to include a `boundBucketClient` property on the `res.locals` object
 // This will allow us to access the BucketClient instance in our route handlers
 // without having to pass it around manually
 declare global {
@@ -178,7 +178,16 @@ app.use((req, res, next) => {
   // Extract the user and company IDs from the request headers
   // You'll want to use a proper authentication and identification
   // mechanism in a real-world application
-  const { user, company } = extractBucketContextFromHeader(req);
+  const user = {
+    id: req.user?.id,
+    name: req.user?.name
+    email: req.user?.email
+  }
+
+  const company = {
+    id: req.user?.companyId
+    name: req.user?.companyName
+  }
 
   // Create a new BoundBucketClient instance by calling the `bindClient` method on a `BucketClient` instance
   // This will create a new instance that is bound to the user/company given.
--- a/packages/node-sdk/README.md
+++ b/packages/node-sdk/README.md
@@ -30,18 +30,12 @@ in **Bucket.co**.
 import { BucketClient } from "@bucketco/node-sdk";
 
 // Create a new instance of the client with the secret key. Additional options
-// are available, such as supplying a logger, fallback features and
-// other custom properties.
-//
-// Fallback features are used in the situation when the server-side
-// features are not yet loaded or there are issues retrieving them.
-// See "Initialization Options" below for more information.
+// are available, such as supplying a logger and other custom properties.
 //
 // We recommend that only one global instance of `client` should be created
 // to avoid multiple round-trips to our servers.
 const client = new BucketClient({
   secretKey: "sec_prod_xxxxxxxxxxxxxxxxxxxxx",
-  fallbackFeatures: ["huddle", "voice-huddle"],
 });
 
 // Initialize the client and begin fetching feature targeting definitions.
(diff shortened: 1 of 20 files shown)


packages/node-sdk/README.md
```diff
@@ -138,6 +132,75 @@ The precedence for configuration options is as follows, listed in the order of i
 2. environment variable
 3. the config file
 
+## Type safe feature flags
+
+To get type checked feature flags, add the list of flags to your `bucket.ts` file.
```
laander, line 137: What's this magical `bucket.ts` file? First mention of it in the README

Since this file isn't required for the consumer to create, let's only mention it as suggestion on how to co-locate bucket specific code

packages/node-sdk/README.md
```diff
@@ -138,6 +132,75 @@ The precedence for configuration options is as follows, listed in the order of i
...
+```
+
+![Type check failed](docs/type-check-failed.png "Type check failed")
+
+## Using with Express
+
+A popular way to integrate the Bucket Node.js SDK is through an express middleware.
+
+```typescript
+import bucket from "./bucket";
```
laander, line 158: Just import directly from the bucket lib?

packages/node-sdk/README.md
```diff
@@ -138,6 +132,75 @@ The precedence for configuration options is as follows, listed in the order of i
...
+import express from "express";
+import { BoundBucketClient } from "../src";
```
laander, line 160: Where does this `src` folder come from? Is it the bucket lib?

packages/node-sdk/README.md
```diff
@@ -138,6 +132,75 @@ The precedence for configuration options is as follows, listed in the order of i
...
+
+// Augment the Express types to include the `boundBucketClient` property on the `res.locals` object
```
laander, line 162:
```suggestion
// Augment the Express types to include a `boundBucketClient` property on the `res.locals` object
```

packages/node-sdk/README.md
```diff
@@ -138,6 +132,75 @@ The precedence for configuration options is as follows, listed in the order of i
...
+    }
+  }
+}
+
+// Add express middleware
+app.use((req, res, next) => {
+  // Extract the user and company IDs from the request headers
+  // You'll want to use a proper authentication and identification
+  // mechanism in a real-world application
+  const { user, company } = extractBucketContextFromHeader(req);
```
laander, line 178: This function is referenced/defined anywhere in the README?
As a reader, I would try to find this function (is it imported? is the defined later in the code example?) which is frustrating/confusing. Can we show an actual working code example here?

packages/node-sdk/README.md
roncohen: no, this is the instance you initialized in a bucket.ts file. I've reference that further up now.

packages/node-sdk/README.md
roncohen: recommended creating a bucket.ts further up now

packages/node-sdk/README.md
roncohen: doh, fixed!

pavkam (approved):

packages/node-sdk/README.md
```diff
@@ -26,28 +26,25 @@ in **Bucket.co**.
...
-});
+export const bucketClient = new BucketClient();
 
 // Initialize the client and begin fetching feature targeting definitions.
 // You must call this method prior to any calls to `getFeatures()`,
 // otherwise an empty object will be returned.
-await client.initialize();
+client.initialize().then({
+  console.log("Bucket initialized!")
+})
```
laander, line 47: what's this `client` ? did you mean `bucketClient` ?

packages/node-sdk/README.md
```diff
@@ -26,28 +26,25 @@ in **Bucket.co**.
 > information that is often sensitive and thus should not be used in
 > client-side applications.
 
+Create a `bucket.ts` file containing the following and set up the
+BUCKET_SECRET_KEY environment variable:
```
laander, line 30: what does this mean? does the `secretKey` automatically try to read from a `BUCKET_SECRET_KEY` env var? if so, I don't think that this is standard practise for 3rd party libs like ours. IMO it's better to be explicit with the passed `secretKey` config instead - all devs know how to pipe in an env var if they want to

packages/node-sdk/README.md
laander: I don't like it :/ I get that this is a "recommended" way to organize your code, but the examples becomes a lot harder to read because of it IMO. i'd much rather we provide a single-file example here that can be read linearly from top to bottom. we can add a note where we suggest placing it into a `bucket.ts` singleton file, but that's not required by the lib

packages/node-sdk/README.md
roncohen: right, doh. Good catch

PR #244, opened 2024-11-11
chore(node-sdk) More docs improvements

roncohen:
Another round of "getting started" instructions.


packages/node-sdk/README.md
```diff
@@ -70,22 +73,6 @@ if (isEnabled) {
...
-  features.huddle?.isEnabled && features.voiceHuddle?.isEnabled;
-```
-
-When using `getFeatures` be careful not to assume that a feature exists, this could be a dangerous pattern:
-
-```ts
-// warning: if the `huddle` feature does not exist because it wasn't created in Bucket
-// or because the client was unable to reach our servers for some reason, this will cause an exception:
-const { isEnabled } = boundClient.getFeatures()["huddle"];
-```
```
roncohen, line 88: getting rid of this to focus on the absolutely necessary

roncohen: brought this back. `getFeatures()` is useful

pavkam (approved):
Left a few minor nitpicks

roncohen:
thanks!

PR #245, opened 2024-11-11
feat(browser-sdk/node-sdk): simplify working with our contexts

roncohen:
We have this annoying situation where it's harder to pass user/company contexts to our SDKs than necessary whenever you _optionally_ have a user/company id.

At the moment, the `id` field in contexts are required: 
Screenshot 2024-11-11 at 15 15 26

However, you often have a situation where you're not sure if you have userId. In the below, `req.user.id` isn't guarenteed and you end up having to jump a bunch of hoops instead of just passing whatever you have:

Screenshot 2024-11-11 at 15 16 25

I liked that the required `id` clearly makes the `id` property/key a built-in, but this is a little frustrating after all. With this change, users can just do the following, and the `user` context will be ignored if `id` isen't set:

Screenshot 2024-11-11 at 15 18 27

Here's the same situation playing out in the React SDK:

Screenshot 2024-11-11 at 15 21 13


packages/browser-sdk/src/context.ts
```diff
@@ -1,15 +1,13 @@
 export interface CompanyContext {
-  id: string | number;
-  [key: string]: string | number;
+  [key: string]: string | number | undefined;
 }
 
 export interface UserContext {
-  id: string | number;
-  [key: string]: string | number;
+  [key: string]: string | number | undefined;
```
pavkam, line 6: could you not mark the `id` here as optional. Same goes to name and email to be explicitly listed but optional.

packages/browser-sdk/src/client.ts
```diff
@@ -101,8 +101,8 @@ export class BucketClient {
     this.logger =
       opts?.logger ?? loggerWithPrefix(quietConsoleLogger, "[Bucket]");
     this.context = {
-      user: opts?.user,
-      company: opts?.company,
+      user: opts?.user?.id ? opts.user : undefined,
```
pavkam, line 104: I guess maybe a test would be good to have for this scenario. I find that these SDKs require a lot of tests due to the nature of js. Same for the node part.

packages/node-sdk/src/types.ts
```diff
@@ -365,13 +365,13 @@ export type TrackOptions = {
  **/
 export type Context = {
   /**
-   * The user context. If the user is set, the user ID is required.
+   * The user context. If no `id` key is set, the whole object is ignored.
    */
-  user?: { id: string; [k: string]: any };
+  user?: Record<string, any>;
```
pavkam, line 370: Same comment as for browser sdk.

packages/node-sdk/src/types.ts
```diff
@@ -365,13 +365,13 @@ export type TrackOptions = {
...
   /**
-   * The company context. If the company is set, the company ID is required.
+   * The company context. If no `id` key is set, the whole object is ignored.
    */
-  company?: { id: string; [k: string]: any };
+  company?: Record<string, any>;
```
pavkam, line 374: do we want to add this to the README?

packages/browser-sdk/src/client.ts
roncohen: good point

pavkam (approved):

roncohen:
this ended up pretty messy for the NodeSDK. I think i need to do another round to simplify tomorrow with fresh eyes

packages/browser-sdk/test/features.test.ts
```diff
@@ -167,7 +166,7 @@ describe("FeaturesClient unit tests", () => {
           isEnabled: true,
           key: "featureB",
           targetingVersion: 1,
-        } satisfies APIFeatureResponse,
```
roncohen, line 170: this never worked, but the tests we're not type checked.

packages/node-sdk/src/client.ts
```diff
@@ -86,7 +86,7 @@ interface ContextWithTracking extends Context {
    * Enable tracking for the context.
    * If set to `false`, tracking will be disabled for the context. Default is `true`.
    */
-  enableTracking: boolean;
+  enableTracking?: boolean;
```
pavkam, line 89: This is technically only needed if users will prepare a context object and pass it around to functions. If they'd call the functions directly with the object notation the interface takes care of the default. But it's better to have it undefined as well for those cases

packages/node-sdk/src/client.ts
```diff
@@ -395,13 +395,8 @@ export class BucketClient {
...
    */
-  private async syncContext({
-    enableTracking = true,
-    ...context
-  }: ContextWithTracking) {
-    checkContextWithTracking({ enableTracking, ...context });
-
-    if (!enableTracking) {
+  private async syncContext(options: ContextWithTracking) {
+    if (!options.enableTracking) {
```
pavkam, line 399: Assuming here that the enable tracking is either true or false and cannot be undefined.

In these cases I love them `assert`s to make this clear to the reader of the code and also find bugs.

packages/node-sdk/src/client.ts
```diff
@@ -501,9 +496,10 @@ export class BucketClient {
    * The company must be set using `withCompany` before calling this method.
    * If the user is set, the company will be associated with the user.
    **/
-  public async updateUser(userId: string, opts?: TrackOptions) {
+  public async updateUser(userId: string | number, opts?: TrackOptions) {
     ok(
-      typeof userId === "string" && userId.length > 0,
+      (typeof userId === "string" && userId.length > 0) ||
+        typeof userId === "number",
```
pavkam, line 502: The error says "companyId" probably my copy paste mistake from the past or something

packages/node-sdk/src/client.ts
```diff
@@ -594,12 +591,13 @@ export class BucketClient {
    * If the company is set, the event will be associated with the company.
    **/
   public async track(
-    userId: string,
+    userId: string | number,
```
pavkam, line 594: I wonder if we should declare User Id and CompanyId as types.

And add two check functions that we can reuse across the code (to check their types)

packages/node-sdk/src/client.ts
```diff
@@ -594,12 +591,13 @@ export class BucketClient {
...
     event: string,
-    opts?: TrackOptions & { companyId?: string },
+    opts?: TrackOptions & { companyId?: string | number },
   ) {
     ok(
-      typeof userId === "string" && userId.length > 0,
+      (typeof userId === "string" && userId.length > 0) ||
+        typeof userId === "number",
```
pavkam, line 600: The errors say userId must be a string while we allow numbers as well. As said before, better to extract this check code into q separate small function and reuse it. Would minimize these types of issues.

packages/node-sdk/src/client.ts
```diff
@@ -657,11 +655,10 @@ export class BucketClient {
     await this._config.batchBuffer.flush();
   }
 
-  private _getFeatures({
-    enableTracking = true,
-    ...context
-  }: ContextWithTracking): Record<string, RawFeature> {
-    void this.syncContext({ enableTracking, ...context });
+  private _getFeatures(
```
pavkam, line 658: Not related to your PR but I really dislike the fact that we have two notations for private members: one uses underscore and the other doesn't. We should use only one. Also I typically document private ones as well so we either do in both case or remove docs from all privates.

packages/node-sdk/src/client.ts
```diff
@@ -657,11 +655,10 @@ export class BucketClient {
...
+    options: ContextWithTracking,
+  ): Record<string, RawFeature> {
+    void this.syncContext(options);
```
pavkam, line 661: Again, assuming that enableTracking is not undefined. So i strongly recommend an assert in sync context (for future developers not to mess up)

packages/node-sdk/src/client.ts
```diff
@@ -762,17 +767,14 @@ export class BucketClient {
...
-          );
+        if (typeof options.user?.id === "undefined") {
+          this._config.logger?.warn("no user set, cannot track event");
           return;
         }
 
-        if (enableTracking) {
-          await this.track(userId, key, {
-            companyId: context.company?.id,
+        if (options.enableTracking) {
```
pavkam, line 775: Same assumption as before. I'd add an assert in this function as well .

packages/node-sdk/src/client.ts
```diff
@@ -793,11 +795,14 @@ export class BucketClient {
     enableTracking = true,
     ...context
   }: ContextWithTracking): TypedFeatures {
-    const features = this._getFeatures({ ...context, enableTracking });
+    const options = { enableTracking, ...context };
+    checkContextWithTracking(options);
```
pavkam, line 799: Can probably be moved to the private getFeatures and shared across all callers. Also safer.

packages/node-sdk/src/client.ts
```diff
@@ -814,17 +819,16 @@ export class BucketClient {
     { enableTracking = true, ...context }: ContextWithTracking,
     key: keyof TypedFeatures,
   ) {
-    const features = this._getFeatures({ ...context, enableTracking });
+    const options = { enableTracking, ...context };
+    checkContextWithTracking(options);
```
pavkam, line 823: Same

packages/node-sdk/src/client.ts
```diff
@@ -845,8 +849,10 @@ export class BucketClient {
       context.company = { id: companyId };
     }
 
-    const contextWithTracking = { ...context, enableTracking: true };
```
pavkam, line 848: Why split then then build again later?

packages/node-sdk/src/client.ts
```diff
@@ -1084,12 +1093,24 @@ export class BoundBucketClient {
 function checkContextWithTracking(context: ContextWithTracking) {
   ok(isObject(context), "context must be an object");
   ok(
-    context.user === undefined || typeof context.user?.id === "string",
-    "user.id must be a string if user is given",
+    typeof context.user === "undefined" || isObject(context.user),
+    "user must be an object if given",
+  );
+  ok(
```
pavkam, line 1099: Same observation. We are repeating this code a lot so better to extract in functions.

packages/node-sdk/src/client.ts
roncohen: OK, adding `checkContextWithTracking` to be sure.

packages/node-sdk/src/client.ts
roncohen: agreed, will cleanup separately.

packages/node-sdk/src/client.ts
roncohen: good catch. Reverting this. Happened just because of back/forth code churn.

packages/node-sdk/src/client.ts
roncohen: extracting

PR #246, opened 2024-11-13
Refactor context checks (follow up from #245)

roncohen:


packages/node-sdk/src/client.ts
```diff
@@ -395,7 +396,7 @@ export class BucketClient {
    *
    * @param context - The context to update.
    */
-  private async syncContext(options: ContextWithTracking) {
+  private async syncContext(options: { enableTracking: boolean } & Context) {
```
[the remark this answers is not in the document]
pavkam: Works as well

packages/node-sdk/src/client.ts
```diff
@@ -1090,28 +1077,26 @@ export class BoundBucketClient {
   }
 }
 
-function checkContextWithTracking(context: ContextWithTracking) {
+function checkContextWithTracking(
+  context: ContextWithTracking,
+): asserts context is ContextWithTracking & { enableTracking: boolean } {
```
pavkam, line 1082: Nice!

packages/node-sdk/src/utils.ts
```diff
@@ -14,6 +14,13 @@ export function ok(condition: boolean, message: string): asserts condition {
   }
 }
 
+export function assertId(id: string | number, entity: string) {
```
pavkam, line 17: Maybe add an optional param `allowUndefined`. Would remove the need for more ifs in code above

pavkam, line 17: Nit: To keep consistent maybe name it idOk

packages/node-sdk/src/client.ts
```diff
@@ -539,13 +536,9 @@ export class BucketClient {
    **/
   public async updateCompany(
     companyId: string | number,
-    opts?: TrackOptions & { userId?: string },
+    opts?: TrackOptions & { userId?: string | number },
```
pavkam, line 539: Still think we can make a type for these instead of repeating it.

pavkam (approved):

packages/node-sdk/src/client.ts
roncohen: done 👍

packages/node-sdk/src/utils.ts
roncohen: renamed!

roncohen: I will leave as it. Hate boolean args ;)

PR #256, opened 2024-11-20
feat(node-sdk): Initialize only once

roncohen:


packages/node-sdk/src/utils.ts
```diff
@@ -160,3 +160,16 @@ export function hashObject(obj: Record<string, any>): string {
 
   return hash.digest("hex");
 }
+
+export function once<T extends () => ReturnType<T>>(
```
pavkam, line 164: Nice to have a test to validate all cases

roncohen: do you mean, it would be nice to have? Aka, please add a test?

pavkam: yes :)

roncohen: don!

pavkam (approved):

PR #263, opened 2024-11-29
docs(browser-sdk): fix: missed  the `requestFeedback` function in one location

roncohen:


packages/browser-sdk/README.md
```diff
@@ -128,6 +128,7 @@ const huddle = bucketClient.getFeature("huddle");
 // {
 //   isEnabled: true,
 //   track: () => Promise<Response>
+//   requestFeedback: () => void
```
laander, line 131: if we're showing the type signature, we should also include the available method arguments (or at least `...` to indicate that "arguments goes here")

laander (approved):

packages/browser-sdk/README.md
roncohen: doh, great point. Updated.

PR #264, opened 2024-11-29
docs(node-sdk): fix json rendering in node-sdk README

roncohen:


packages/node-sdk/README.md
```diff
@@ -127,14 +127,14 @@ Note: BUCKET_FEATURES_ENABLED, BUCKET_FEATURES_DISABLED are comma separated list
...
-  host: "https://proxy.slick-demo.com"
-  featureOverrides: {
-    huddles: true,
-    voiceChat: false
-  },
+  "secretKey": "...",
+  "logLevel": "warn",
+  "offline": true,
+  "host": "https://proxy.slick-demo.com",
+  "featureOverrides": {
```
pavkam, line 134: `featureOverrides` is declared as follows:
```
featureOverrides?:
    | string
    | ((context: Context) => Partial<Record<keyof TypedFeatures, boolean>>);
```
so you should use an arrow function

roncohen: this is the config file which takes a simple boolean map

pavkam: Ah!

pavkam (approved):

PR #265, opened 2024-12-06
add autoplacement middleware

laander:
Positioning the widget in a popover state can be tricky given the various environments that it can used in. With the current positioning where `bottom` is preferred, it can result in very crowded placements if there's very little room for it.

Before, example:
image

Using the `flip` middleware for floating UI, it will flip to another side when there's not enough room available in the preferred direction. It will prefer following placements, in this order: `bottom`, `top`, `right`, `left`.

After, example:
image


Swiftwork (approved):
Conceptually and code wise it looks fine, if it works ^^

PR #270, opened 2024-12-17
chore: updating playwright

roncohen:
Tests are currently failing due to an updated base image which no longer
contains the playwright dependencies:
https://github.com/bucketco/bucket-javascript-sdk/actions/runs/12380013335/job/34555430888


pavkam (approved):
✈️

PR #273, opened 2024-12-20
Improve handling of contexts

pavkam:
This PR does the following:
- Fixes the handling of empty contexts in some situations, and makes sure that no erroneous requests are sent to the API,
- Removes an internal field from being added to the context,
- A rate-limited warning is added when evaluating features complaining about missing context fields. This should help developers detect that their contexts are lacking details,
- A "defensive" check for the status code of the HTTP response, even if it might not be needed.


packages/node-sdk/src/client.ts
```diff
@@ -271,7 +277,12 @@ export class BucketClient {
 
       this._config.logger?.debug(`get request to "${url}"`, response);
 
-      if (!isObject(response.body) || response.body.success !== true) {
+      if (
+        !isObject(response.body) ||
+        response.body.success !== true ||
+        response.status < 200 ||
+        response.status >= 400
```
roncohen, line 284: I think we can just use `response.ok` instead of these last two

packages/node-sdk/src/client.ts
```diff
@@ -849,16 +903,24 @@ export class BucketClient {
     };
     checkContextWithTracking(contextWithTracking);
 
-    const params = new URLSearchParams(flattenJSON({ context }));
+    const params = new URLSearchParams(
+      Object.keys(context).length ? flattenJSON({ context }) : undefined,
```
roncohen, line 907: first line in `flattenJSON` is `if (Object.keys(data).length === 0) return {};`. This doesn't seem necessary?

packages/node-sdk/src/client.ts
```diff
@@ -849,16 +903,24 @@ export class BucketClient {
...
+    );
+
     if (key) {
       params.append("key", key);
     }
+    const paramsString = params.size ? `?${params.toString()}` : "";
```
roncohen, line 913: what does this fix? an empty URLSearchParams returns empty string.

packages/node-sdk/src/client.ts
```diff
@@ -234,9 +234,15 @@ export class BucketClient {
...
       >(url, this._config.headers, body);
+
       this._config.logger?.debug(`post request to "${url}"`, response);
 
-      if (!isObject(response.body) || response.body.success !== true) {
+      if (
+        !isObject(response.body) ||
+        response.body.success !== true ||
+        response.status < 200 ||
+        response.status >= 400
```
roncohen, line 244: I think we can just use response.ok instead of these last two

roncohen (approved):
generally good improvements 👍

packages/node-sdk/src/client.ts
pavkam: Makes sense, would need to expose it.

packages/node-sdk/src/client.ts
pavkam: That's the issue actually. An empty object serializes as "[Object]" string which is not what we want.

packages/node-sdk/src/client.ts
pavkam: Just wanted to avoid having a ? suffix when there's no params. Probably overkill though.

packages/node-sdk/src/client.ts
pavkam: Will do

packages/node-sdk/src/client.ts
roncohen:
```
❯ node
Welcome to Node.js v18.16.0.
Type ".help" for more information.
> new URLSearchParams({}).toString()
''
>
```

packages/node-sdk/src/client.ts
roncohen: expose it?

packages/node-sdk/src/client.ts
pavkam: Maybe because it's an empty object within an empty object. I get: `"https:// api. example. com/ features/ evaluated?context=%5Bobject+Object%5D"`

PR #274, opened 2024-12-24
Refactor SDK to use updated `evaluateFeatureRules` API

pavkam:
Updated the Node and Browser SDKs to utilize the new `evaluateFeatureRules` method from `@bucketco/flag-evaluation@~1.0.0`. This enhances the clarity and compatibility of feature evaluation logic. Adjusted related test cases and bumped package version to 1.4.4.

### DIFF
--- a/packages/browser-sdk/package.json
+++ b/packages/browser-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/browser-sdk",
-  "version": "2.4.1",
+  "version": "2.4.2",
   "packageManager": "yarn@4.1.1",
   "license": "MIT",
   "repository": {
--- a/packages/flag-evaluation/package.json
+++ b/packages/flag-evaluation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/flag-evaluation",
-  "version": "1.0.0",
+  "version": "0.1.0",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/packages/flag-evaluation/package.json
+++ b/packages/flag-evaluation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/flag-evaluation",
-  "version": "0.0.7",
+  "version": "1.0.0",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/packages/flag-evaluation/src/index.ts
+++ b/packages/flag-evaluation/src/index.ts
@@ -1,4 +1,3 @@
-import assert from "node:assert";
 import { createHash } from "node:crypto";
 
 /**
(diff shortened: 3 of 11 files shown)


packages/flag-evaluation/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/flag-evaluation",
-  "version": "0.0.7",
+  "version": "1.0.0",
```
roncohen, line 3: let's keep at `0.x.x`.. In semver, different rules apply below the big 1.0 and I'd like to signal that this library isn't for general consumption. Lets make this `0.1.0`

packages/node-sdk/src/client.ts
```diff
@@ -650,6 +412,341 @@ export class BucketClient {
...
+          }
+
+          return res;
+        },
+      );
+    }
+
+    return this._config.featuresCache;
+  }
+
```
roncohen, line 749: since this was moved, it's hard to know if anything actually changed :/

packages/flag-evaluation/test/index.test.ts
```diff
@@ -365,3 +358,244 @@ describe("rollout hash", () => {
     });
   }
 });
+
+describe("flattenJSON", () => {
```
roncohen, line 362: nice!

packages/browser-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/browser-sdk",
-  "version": "2.4.1",
+  "version": "2.4.2",
```
roncohen, line 3: browser-sdk doesn't use the flag-evaluation package, but i think it's fine to do a new patch release since we moved a bunch of stuff around in here.

roncohen (approved):
only minor comments. Good work on adding tests and comments!

packages/flag-evaluation/src/index.ts
```diff
@@ -106,49 +226,40 @@ export function flattenJSON(data: object): Record<string, string> {
...
+ * @return {Record<string, any>} The unflattened JSON object with nested structure restored.
+ */
+export function unflattenJSON(data: Record<string, any>): Record<string, any> {
   const result: Record<string, any> = {};
 
   for (const i in data) {
     const keys = i.split(".");
     keys.reduce((acc, key, index) => {
       if (index === keys.length - 1) {
+        assert(typeof acc === "object", "overlapping keys");
```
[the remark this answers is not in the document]
pavkam: It will explodat anyway even without this assert because we would try to set a property on a string and that will trigger an error. Should I simply skip this assignment in this case to make it not throw?

packages/node-sdk/src/client.ts
pavkam: Nothing changed there and sorry, my IDE wants to be smart and group things by visibility. Might want to tweak that so it's not doing that.

packages/browser-sdk/package.json
pavkam: That was my thinking as well.

packages/flag-evaluation/src/index.ts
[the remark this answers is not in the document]
roncohen: ah, i see. Yeah, let's simply skip it 👍

PR #275, opened 2025-01-02
Introduce methods to update context after intialization

roncohen:
This introduced methods to let you update the user/company/other context after initialization. This lets you easily update data when it changes without needing to re-initialize etc. Useful for example to let users self opt-in to experimental features.

The PR also introduces some plumming needed to make sure that any changes that happen to feature targeting after you update the user/company/other is reflected, live, in the `useFeature` calls you have in your code.

### DIFF
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -193,7 +193,7 @@ export class BucketClient {
   /**
    * Update the user context.
    * @description Performs a shallow merge with the existing user context.
-   * Updates to the user ID will be ignored.
+   * Attempting to update the user ID will log a warning and be ignored.
    *
    * @param user
    */
@@ -217,7 +217,7 @@ export class BucketClient {
   /**
    * Update the company context.
    * Performs a shallow merge with the existing company context.
-   * Updates to the company ID will be ignored.
+   * Attempting to update the company ID will log a warning and be ignored.
    *
    * @param company
    */
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -242,6 +242,87 @@ export class BucketClient {
   }
 
   /**
+   * Update the user context.
+   * @description Performs a shallow merge with the existing user context.
+   * Updates to the user ID will be ignored.
+   *
+   * @param user
+   */
+  public async updateUser(user: {
+    [key: string]: string | number | undefined;
+  }) {
+    if (user.id && user.id !== this.context.user?.id) {
+      this.logger.warn(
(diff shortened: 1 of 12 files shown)


packages/browser-sdk/src/client.ts
```diff
@@ -190,6 +190,83 @@ export class BucketClient {
...
+
+  /**
+   * Update the company context.
+   * Performs a shallow merge with the existing company context.
+   * Updates to the company ID will be ignored.
+   *
+   * @param company
+   */
+  async updateCompany(company: { [key: string]: string | number | undefined }) {
+    if (company.id && company.id !== this.context.company?.id) {
```
pavkam, line 225: This check would refuse to update the company if the existing context doesn't have a company set, since the new company ID would be defined and the existing one would be `undefined`.

packages/browser-sdk/src/feature/features.ts
```diff
@@ -83,6 +85,12 @@ export interface CheckEvent {
   version?: number;
 }
 
+type context = {
```
pavkam, line 88: Why not use `BucketContext`?

packages/react-sdk/src/index.tsx
```diff
@@ -115,18 +115,19 @@ export function BucketProvider({
       sdkVersion: SDK_VERSION,
     });
     clientRef.current = client;
+
+    client.onFeaturesUpdated(() => {
```
pavkam, line 119: Can we provide an effect hook for this `useFeaturesUpdatedEffect()`?

pavkam (commented):
Looks good, left a few comments, and are we missing a README update?

packages/react-sdk/src/index.tsx
roncohen: hmm, yes. What's the use-case?

packages/browser-sdk/src/client.ts
pavkam: It's confusing because there's `id: company.id ?? this.context.company?.id` on line `234` which made me think that is an option. Maybe also update the comment to say "attempts to update the company ID will result in the call being discarded"? Otherwise it sounds like the company ID is ignored, but not the other parts.

packages/react-sdk/src/index.tsx
pavkam: Oh, ignore me. I thought this change was in an example. It's actually in the SDK itself.

pavkam (approved):

packages/browser-sdk/src/client.ts
roncohen: gotcha, makes sense

PR #276, opened 2025-01-02
Update the browser SDK to use `baseUrl`s instead of `host`s

pavkam:
This PR, changes the naming of the `host` and `sseHost` parameters to better reflect them being `base urls`. Also, fixed the handling of SSE URL creation.

### DIFF
--- a/packages/browser-sdk/README.md
+++ b/packages/browser-sdk/README.md
@@ -86,19 +86,19 @@ Supply these to the constructor call:
 
 ```ts
 type Configuration = {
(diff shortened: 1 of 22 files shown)


roncohen (approved):
nice!

roncohen:
we should make the terminology changes (host -> apiBaseUrl) for the ReactSDK as well and update the docs

pavkam:
> we should make the terminology changes (host -> apiBaseUrl) for the ReactSDK as well and update the docs

I'll make a follow-up PR

PR #277, opened 2025-01-02
Update SDKs to use `apiBaseUrl` instead of `host`

pavkam:
Replaced the `host` property with `apiBaseUrl` across all SDKs for improved clarity and consistency. Updated tests, documentation, and implementation to reflect the change while marking `host` as deprecated for backward compatibility. Bumped versions to indicate non-breaking minor changes.


roncohen (approved):
nice! thanks!

PR #286, opened 2025-01-17
feat(browser-sdk): support overrides, propagate updates

roncohen:
This is a piece of infrastructure needed for local overrides. We will likely need a few adjustments to this as we go about building the toolbar on top, but this is a good base.

Fixes BUC-3235

### DIFF
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -498,6 +498,8 @@ export class BucketClient {
   /**
    * Returns a map of enabled features.
    * Accessing a feature will *not* send a check event
+   * and `isEnabled` does not take any feature overrides
+   * into account.
    *
    * @returns Map of features
    */
@@ -513,13 +515,13 @@ export class BucketClient {
     const f = this.getFeatures()[key];
 
     const fClient = this.featuresClient;
-    const value = f?.isEnabled ?? false;
+    const value = f?.isEnabledOverride ?? f?.isEnabled ?? false;
 
     return {
       get isEnabled() {
         fClient
           .sendCheckEvent({
-            key: key,
+            key,
             version: f?.targetingVersion,
             value,
           })
@@ -538,6 +540,14 @@ export class BucketClient {
         });
       },
     };
+  }
+
+  setFeatureOverride(key: string, isEnabled: boolean | null) {
+    this.featuresClient.setFeatureOverride(key, isEnabled);
+  }
+
+  getFeatureOverride(key: string): boolean | null {
+    return this.featuresClient.getFeatureOverride(key);
   }
 
   sendCheckEvent(checkEvent: CheckEvent) {
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -198,13 +198,13 @@ export class BucketClient {
   private context: BucketContext;
   private config: Config;
   private requestFeedbackOptions: Partial<RequestFeedbackOptions>;
-  private logger: Logger;
   private httpClient: HttpClient;
 
   private autoFeedback: AutoFeedback | undefined;
   private autoFeedbackInit: Promise<void> | undefined;
   private featuresClient: FeaturesClient;
 
+  public readonly logger: Logger;
   /**
    * Create a new BucketClient instance.
    */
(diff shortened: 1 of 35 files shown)


packages/browser-sdk/src/feature/features.ts
```diff
@@ -138,14 +145,38 @@ type context = {
 
 export const FEATURES_EXPIRE_MS = 30 * 24 * 60 * 60 * 1000; // expire entirely after 30 days
 
-const localStorageCacheKey = `__bucket_features`;
+const localStorageFetchedFeaturesKey = `__bucket_fetched_features`;
+const localStorageOverridesKey = `__bucket_overrides`;
+
+function setOverridesCache(overrides: Record<string, boolean>) {
```
laander, line 151: Let's pull the overrides type out into a reusable type of it's own since it's used multiple times
`export type OverridesFeatures = Record<string, boolean>`

packages/browser-sdk/src/feature/features.ts
```diff
@@ -138,14 +145,38 @@ type context = {
...
+  localStorage.setItem(localStorageOverridesKey, JSON.stringify(overrides));
+}
+
+function getOverridesCache(): Record<string, boolean> {
```
laander, line 155: Use reusable type

packages/browser-sdk/src/feature/features.ts
```diff
@@ -138,14 +145,38 @@ type context = {
...
+}
 
 /**
  * @internal
  */
 export class FeaturesClient {
   private cache: FeatureCache;
-  private features: RawFeatures;
+  private fetchedFeatures: FetchedFeatures;
+  private featureOverrides: Record<string, boolean>;
```
laander, line 176: Use reusable type

packages/browser-sdk/src/feature/features.ts
```diff
@@ -291,11 +324,41 @@ export class FeaturesClient {
...
+    // add any overrides that aren't in the fetched features
+    for (const key in this.featureOverrides) {
+      if (!this.features[key]) {
+        mergedFeatures[key] = {
+          key,
+          isEnabled: false,
+          isEnabledOverride: null,
+        };
+      }
+    }
```
laander, line 350: why is this useful? especially when `isEnabled` and `isEnabledOverride` is always set to `false` and `null` instead of the actual overriden value?

packages/browser-sdk/src/feature/features.ts
```diff
@@ -28,7 +28,14 @@ export type RawFeature = {
 
 const FEATURES_UPDATED_EVENT = "features-updated";
 
-export type RawFeatures = Record<string, RawFeature | undefined>;
+export type FetchedFeatures = Record<string, FetchedFeature | undefined>;
+export type RawFeature = FetchedFeature & {
```
laander, line 32: can we rename this? confuses me slightly as a `raw` indicates that it's not manipulated internally, whereas overrides is a form of manipulation.

perhaps `LocalFeature` ?

packages/browser-sdk/src/feature/features.ts
roncohen: > isEnabledOverride is always set to false and null instead of the actual overriden value?

that's a bug. Thanks!

fetched features is what the API returns. If you've added a feature locally but not on app.bucket.co we should still include it.

packages/browser-sdk/src/feature/features.ts
roncohen: this is unfortunately already exported so renaming would be a breaking change. Added a note

packages/browser-sdk/src/feature/features.ts
laander: right 👍

laander (approved):

packages/browser-sdk/src/client.ts
```diff
@@ -513,13 +515,13 @@ export class BucketClient {
     const f = this.getFeatures()[key];
 
     const fClient = this.featuresClient;
-    const value = f?.isEnabled ?? false;
+    const value = f?.isEnabledOverride ?? f?.isEnabled ?? false;
```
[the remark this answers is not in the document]
roncohen: preliminary toolbar code uses `bucketClient.getFeatures()` (plural) which has all the data needed. In any case we can adjust once this is in 👍

PR #290, opened 2025-01-20
Toolbar (alpha)

roncohen:
The toolbar enables teams to quickly enable/disable features directly in their own app.

It's a UI that renders on top of the page:
image

Features so far:
- fetches features remotely through the SDK
- also takes a list of features passed locally by prop
- override feature enabled easily by flipping a switch, which is stored in localstorage
- filter the list of features by search term
- works for both Browser SDK and React SDK

This started out as a way to see just how hard it would be to implement a toolbar for Bucket, so the code is now really in great shape.

Known issues:
- we use `featureList` to pass the string[] with feature keys into the BuckerProvider (and from there to BucketClient from the browser-sdk because `features` was already take for misc. feature fetching related options
- positioning (bottom-left etc.) of the toolbar toggle doesn't really work the the toolbar dialog open up weirdly/wonkily in lots of situation (like narrow screen).


roncohen:
There's both `index.html` and `example/browser.html` at the moment which threw me off. Let's kill one of them? Otherwise LGTM. I can't approve because it's my PR, but if you approve it you should be able to also merge it

laander:
Bundle size differences.

**Browser SDK**

Before:
image

After:
image

**React SDK**

Before:
image

After:
image

packages/browser-sdk/src/toolbar/Switch.tsx
```diff
@@ -0,0 +1,50 @@
+import { Fragment, h } from "preact";
+
+interface SwitchProps extends h.JSX.HTMLAttributes<HTMLInputElement> {
+  isOn: boolean;
```
Swiftwork, line 4: I would probably call this `checked` to match HTML and Chakra (v3) naming-conventions.

Swiftwork (approved):
A small nit-pick only. There were a lot of changes but nothing jumped to me as crazy. The best is to test it locally for a while and discover bugs that way. Have you tested it in safari, chrome, firefox, and maybe edge?

packages/browser-sdk/src/toolbar/Switch.tsx
laander: makes sense to rename it, done 👍

PR #293, opened 2025-01-24
feat(openfeature-*-provider): make use of the remote config

pavkam:
This PR introduces functionality into the `OpenFeature` SDKs to make use of the `config` objects exposed by the primary SDKs.

Also, the PR fixes a number of inconsistencies across the SDKs and updates the documentation accordingly.


packages/openfeature-browser-provider/src/index.ts
```diff
@@ -100,66 +100,96 @@ export class BucketBrowserSDKProvider implements Provider {
...
-      } satisfies ResolutionDetails<boolean>;
+        errorMessage: "Bucket client not initialized",
+      } satisfies ResolutionDetails<T>;
+    }
 
     const features = this._client.getFeatures();
     if (flagKey in features) {
       const feature = this._client.getFeature(flagKey);
+
+      if (!feature.isEnabled) {
```
pavkam, line 122: @roncohen said in another PR:
```
looks like this will return default value is the feature is not enabled? When fetching configs, i think we should just return the config value regardless of the isEnabled state. If they want this logic, they'll need to handle that in their application.
```

Putting here for discussion.

roncohen:
could you update the base branch for this so the "Changed" view only contains the stuff from this PR?

packages/openfeature-browser-provider/src/index.ts
pavkam: I have refactored the code a tad to split the `disabled` from other logic, but still, it needs more looking.

roncohen (approved):
looks good. Thanks for improving and reorganizing the tests!

before merging: could you update the `example` apps (both node and browser) to include an example of the usage of Remote config so we can demonstrate that it works in a real work app? thanks!

PR #304, opened 2025-02-03
chore: auto-publish in "v*" branches

roncohen:
Since `main` has gone to 1.6-alpha, we need to start a separate branch for 1.5 series releases. This'll let us run the publish script from any branch that starts with `v` as in `v1.5`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - v*
 
 jobs:
   release:


laander (approved):

roncohen:
> Should we consider using tags for realses?

definitely! And I now realize this branching strategy isn't going to work because each package is independently versioned.

laander:
hmm, come to think of it, we can't really talk about the whole repo being at a specific version since it's separate packages e.g. react-sdk is at `v3.0.0 -alpha` and node-sdk is at `v1.6.0-alpha`..

branch names could start with a version, like `v1.5` but should include the SDK in question to reduce confusion, like `v1.5-node-sdk`

[this branch](https://github.com/bucketco/bucket-javascript-sdk/tree/1.5) for example is too generic / unspecific

PR #307, opened 2025-02-05
feat(node-sdk): improved flag events

pavkam:
This PR improves the handling of check/evaluate events in the age of `remote config`. Fixes some additional small nagging issues along the way.


matus-vacula (approved):

packages/browser-sdk/README.md
```diff
@@ -172,6 +174,43 @@ by down-stream clients, like the React SDK.
...
+const unsubscribe = bucketClient.onFeaturesUpdated(() => {
+  console.log("Features were updated");
+});
+
+// Later, stop listening for updates
+unsubscribe();
+```
+
+> [!NOTE]
+> Note that the callback may be called even if features haven't actually changed.
```
roncohen, line 212: I don't think these notes work on GitBook unfortunately :/

pavkam: Sucks :( Can't find something in common.

pavkam: And pedantic I went... added an awk script to convert from GitHub to GitBook format so now we can use GitHub format and it'll look good in both places.

roncohen: hah, nice!

PR #308, opened 2025-02-05
Remove deprecated and flatten options

roncohen:
- standardize on `features` for passing a feaure list
- take arguments from the old `features` options and put them in
  at the root options map.


packages/browser-sdk/src/client.ts
```diff
@@ -235,10 +229,32 @@ export interface InitOptions {
...
 
   /**
-   * @deprecated
-   * Use `sseBaseUrl` instead.
+   * Feature keys for which `isEnabled` should fallback to true
+   * if SDK fails to fetch features from Bucket servers. If a record
+   * is supplied instead of array, the values of each key represent the
+   * configuration values and `isEnabled` is assume `true`.
+   */
+  fallbackFeatures?: string[] | Record<string, FallbackFeatureOverride>;
```
laander, line 237: Let's get this merged into `features` too? 
e.g. 
```ts 
{ key: "my-foo-feature", fallback: true | false }
```

roncohen: sure, but I will do it separately. Still need to figure out how the fallback stuff is going to work, potentially syncing it down from bucket.co.

laander (approved):

PR #309, opened 2025-02-06
fix(browser-sdk,react-sdk): remove featureId option for requestFeedback

roncohen:
Remove legacy `featureId` option when using the `requestFeedback` and `feedback` methods

The actual feedback payload can still contain the featureId because bucket's prompting message do not include the feature key.


pavkam (commented):
Just a reminder: This will be a breaking change. Maybe get rid of all deprecated fields in this PR so we don't have to make more of them?

roncohen:
I removed deprecated options related to feedback. I didn't find any other deprecated options in the browser/react SDKs.

PR #311, opened 2025-02-08
feat(node-sdk): add configurable exit event flushing

pavkam:
This PR Introduces `flushOnExit` option in batch buffer configuration. When on (default), the client will flush all events on it's exit.

### DIFF
--- a/packages/node-sdk/README.md
+++ b/packages/node-sdk/README.md
@@ -330,17 +330,6 @@ const client = new BucketClient({
 > If you are creating multiple client instances in your application, it's recommended to disable `flushOnExit`
 > to avoid potential conflicts during process shutdown. In such cases, you should implement your own flush handling.
 
-A naive example of manual flush handling:
-
-```typescript
-process.on("SIGINT", () => {
-  console.log("Flushing batch buffer...");
-  client.flush().then(() => {
-    process.exit(0);
-  });
-});
-```
-
 When you bind a client to a user/company, this data is matched against the
 targeting rules. To get accurate targeting, you must ensure that the user/company
 information provided is sufficient to match against the targeting rules you've
--- a/packages/node-sdk/README.md
+++ b/packages/node-sdk/README.md
@@ -315,7 +315,22 @@ the number of calls that are sent to Bucket's servers. During process shutdown,
 some messages could be waiting to be sent, and thus, would be discarded if the
 buffer is not flushed.
 
-A naive example:
+By default, the SDK automatically subscribes to process exit signals and attempts to flush
+any pending events. This behavior is controlled by the `flushOnExit` option in the client configuration:
+
+```typescript
+const client = new BucketClient({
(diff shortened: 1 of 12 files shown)


roncohen:
Got this error on the first try:

Screenshot 2025-02-09 at 16 58 56

pavkam:
> Got this error on the first try:
> Screenshot 2025-02-09 at 16 58 56

Yeah, me, trying to be smart, removed the anonymous function wrapper, and `this` has disappeared. Pushed a commit

packages/node-sdk/src/flusher.ts
```diff
@@ -0,0 +1,64 @@
...
+  killSignals.forEach((signal) => {
+    const hasListeners = process.listenerCount(signal) > 0;
+
+    if (hasListeners) {
+      process.prependListener(signal, wrappedCallback);
+    } else {
+      process.on(signal, async () => {
+        await wrappedCallback();
+        process.exit(0x100 + constants.signals[signal]);
+      });
```
roncohen, line 52: I tried running a process with and without this handler to ensure that the exit codes stay the same. 

This should be `process.exit(128 + constants.signals[signal]);` as far as i can tell.

packages/node-sdk/src/flusher.ts
```diff
@@ -0,0 +1,64 @@
...
+      return;
+    }
+
+    state = false;
+
+    try {
+      const result = await Promise.race([
+        new Promise((resolve) => setTimeout(() => resolve(true), timeout)),
+        callback(),
+      ]);
```
roncohen, line 26: try running this program:
```
import client from "./bucket";

async function start() {
  await client.initialize();

  client.getFeature({}, "create-todos");

  console.log("done");
}

start();
```

you'll see that these lines add an additional 5s before existing

packages/node-sdk/src/flusher.ts
pavkam: Getting old. Ox100 is 256 not 128 🤦

packages/node-sdk/src/flusher.ts
roncohen: check out `setTimeout().unref()`

packages/node-sdk/README.md
```diff
@@ -315,7 +315,22 @@ the number of calls that are sent to Bucket's servers. During process shutdown,
 some messages could be waiting to be sent, and thus, would be discarded if the
 buffer is not flushed.
 
-A naive example:
+By default, the SDK automatically subscribes to process exit signals and attempts to flush
```
roncohen, line 318: above we say

> It is highly recommended that users of this SDK manually call `flush()`
method on process shutdown.

i don't think we need to recommend that anymore. This auto-flush should "just work".

roncohen (approved):
looks good after the current comments have been resolved 👍

PR #312, opened 2025-02-08
feat(node-sdk): allow passing `meta` to `getFeature(s)`

pavkam:
This PR adds the ability to pass the `meta` to `getFeature(s)` and the bound client.

### DIFF
--- a/packages/node-sdk/package.json
+++ b/packages/node-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/node-sdk",
-  "version": "1.5.1",
+  "version": "1.5.3",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/packages/node-sdk/package.json
+++ b/packages/node-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/node-sdk",
-  "version": "1.5.3",
+  "version": "1.5.2",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/packages/node-sdk/src/client.ts
+++ b/packages/node-sdk/src/client.ts
@@ -124,6 +124,7 @@ export class BucketClient {
    * See README for configuration options.
    *
    * @param options - The options for the client or an existing client to clone.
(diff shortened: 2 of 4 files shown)


packages/node-sdk/src/client.ts
```diff
@@ -114,6 +124,16 @@ export class BucketClient {
...
    * @param options - The options for the client or an existing client to clone.
+   * @param options.apiBaseUrl - The base URL to send requests to (optional).
+   * @param options.logger - The logger to use for logging (optional).
+   * @param options.httpClient - The HTTP client to use for sending requests (optional).
+   * @param options.logLevel - The log level to use for logging (optional).
+   * @param options.offline - Whether to run in offline mode (optional).
+   * @param options.fallbackFeatures - The fallback features to use if the feature is not found (optional).
+   * @param options.batchOptions - The options for the batch buffer (optional).
+   * @param options.featureOverrides - The feature overrides to use for the client (optional).
+   * @param options.configFile - The path to the config file (optional).
```
roncohen, line 135: `secretKey` missing in the comments?

packages/node-sdk/src/client.ts
```diff
@@ -987,6 +1053,14 @@ export class BoundBucketClient {
   private readonly _client: BucketClient;
   private readonly _options: ContextWithTracking;
 
+  /**
+   * Creates a new BoundBucketClient.
```
roncohen, line 1057: this is basically an "internal" constructor. You should be using the `client.bindClient()` method and not this one directly. Can we make it `@internal`?

roncohen (approved):
few nits, but otherwise LGTM!

packages/node-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/node-sdk",
-  "version": "1.5.1",
+  "version": "1.5.3",
```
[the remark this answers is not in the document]
pavkam: Haha, there's another PR with 1.5.2 so I went to 1.5.3 directly. That being said, we really have to figure out a better way to separate "work" from "releases".

PR #325, opened 2025-02-15
feat(browser-sdk): Event listeners

roncohen:
Hooks allow capturing different events that happen in the Browser SDK.

They are useful for a number of things like building client side integrations to analytics or error-logging systems.

### DIFF
--- a/packages/browser-sdk/README.md
+++ b/packages/browser-sdk/README.md
@@ -322,7 +322,7 @@ Hooks allow for capturing various events occurring in the BucketClient. There ar
 Supply a list of `Hook`s in the BucketClient constructor or use the `on()` method to add a hook after construction. See the API reference for details on each hook.
 
 ```ts
-import { BucketClient, CheckEvent } from "@bucketco/browser-sdk";
+import { BucketClient, CheckEvent, RawFeatures } from "@bucketco/browser-sdk";
 
 const client = new BucketClient({
   // other options...
--- a/packages/browser-sdk/README.md
+++ b/packages/browser-sdk/README.md
@@ -109,7 +109,6 @@ type Configuration = {
     staleTimeMs?: number; // at initialization time features are loaded from the cache unless they have gone stale. Defaults to 0 which means the cache is disabled. Increase in the case of a non-SPA
     expireTimeMs?: number; // In case we're unable to fetch features from Bucket, cached/stale features will be used instead until they expire after `expireTimeMs`. Default is 30 days
   };
-  hooks?: Hooks[] | Hooks[][]; // See the "Hooks" section below.
 };
 ```
 
@@ -311,7 +310,7 @@ See details in [Feedback HTTP API](https://docs.bucket.co/reference/http-trackin
 
 ### Hooks
 
-Hooks allow for capturing various events occurring in the BucketClient. There are 5 kinds of events:
+Hooks allow for capturing various events occurring in the `BucketClient`. There are 5 kinds of events:
 
 - FeaturesUpdated
 - User
@@ -319,20 +318,13 @@ Hooks allow for capturing various events occurring in the BucketClient. There ar
 - Check
 - Track
 
-Supply a list of `Hook`s in the BucketClient constructor or use the `on()` method to add a hook after construction. See the API reference for details on each hook.
+Use the `on()` method to add a hook after construction. See the API reference for details on each hook.
 
 ```ts
 import { BucketClient, CheckEvent, RawFeatures } from "@bucketco/browser-sdk";
 
 const client = new BucketClient({
-  // other options...
-  hooks: [
-    {
-      type: "features-updated",
-      handler: (features: RawFeatures) =>
-        console.log(`features updated: ${features}`),
-    },
-  ],
+  // regular options
 });
 
 // or add the hooks after construction:
--- a/packages/browser-sdk/README.md
+++ b/packages/browser-sdk/README.md
@@ -328,7 +328,7 @@ const client = new BucketClient({
 });
 
 // or add the hooks after construction:
-const unsub = client.on("isEnabledCheck", (check: CheckEvent) =>
+const unsub = client.on("enabledCheck", (check: CheckEvent) =>
   console.log(`Check event ${check}`),
 );
 // use the returned function to unsubscribe, or call `off()` with the same arguments again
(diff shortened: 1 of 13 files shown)


packages/browser-sdk/src/hooksManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    "check-config": [],
+    "check-is-enabled": [],
+    "features-updated": [],
+    user: [],
+    company: [],
+    track: [],
+  };
+
+  addHook(hook: Hook): void {
+    (this.hooks[hook.type] as Hook[]).push(hook);
```
pavkam, line 83: Think it's "safer" to add the type on the field itself rather than casting here

packages/browser-sdk/src/hooksManager.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Hook for features updated event.
+ */
+export type FeaturesUpdatedHook = {
+  type: "features-updated";
+  callback: (features: RawFeatures) => void;
+};
+
+type trackEvent = {
```
pavkam, line 36: Nit: types should be PascalCased even if not exported

packages/browser-sdk/src/hooksManager.ts
```diff
@@ -0,0 +1,105 @@
...
+    user: [],
+    company: [],
+    track: [],
+  };
+
+  addHook(hook: Hook): void {
+    (this.hooks[hook.type] as Hook[]).push(hook);
+  }
+
+  triggerCheck(checkEvent: CheckEvent): void {
```
pavkam, line 86: I think you only need one function to trigger them: trigger(name: Hook["type"], payload: Hook[name]) ... something like this would add the type checking

packages/browser-sdk/README.md
```diff
@@ -308,6 +309,18 @@ If you are not using the Bucket Browser SDK, you can still submit feedback using
 
 See details in [Feedback HTTP API](https://docs.bucket.co/reference/http-tracking-api#feedback)
 
+### Hooks
+
+Hooks allow for capturing various events occurring in the BucketClient. There are 5 kinds of events:
+
+- FeaturesUpdated
```
pavkam, line 316: A bit more docs would be a plus. And an example

packages/browser-sdk/src/hooksManager.ts
roncohen: `trigger(name: Hook["type"], payload: Hook[name])` doesn't work, and I didn't find a way to make those types work unfortunately. If you have ideas I'm open for it!

packages/browser-sdk/src/hooksManager.ts
pavkam: I meant to make this.hooks be of type Hooks[] in the definition

packages/browser-sdk/src/hooksManager.ts
pavkam: There's a couple of ways. One is to use Extract<Hook, { type }>. This should give you the union member that matches that given piece of the type (where type is your type parameter. I think your biggest issue is that type is not predefined so typescript sees it as 'string' instead of an union of those possible string costs

roncohen: Updated, let me know what you think!

packages/browser-sdk/README.md
roncohen: done!

packages/browser-sdk/src/hooksManager.ts
roncohen: fixed!

packages/browser-sdk/src/hooksManager.ts
roncohen: updated. There's still some casting going on. I've been back and forth and it's tricky to get rid of the casts. Try pulling down the code and have a go at it if you have an idea.

Swiftwork (changes requested):
Code looks good but I would like to voice my opinion that I think you should only have the `.on` paradigm and not the `hooks = []`. The `.on` offers in my opinion enough flexibility and makes it more clear how you use it. Less is more kind of thinking.

There should also be an `.off` method or the `.on` method should return an "unsubscribe" function. This is important to avoid "leaks" and allow rerender without duplicating the listener.

roncohen:
> Code looks good but I would like to voice my opinion that I think you should only have the `.on` paradigm and not the `hooks = []`. The `.on` offers in my opinion enough flexibility and makes it more clear how you use it. Less is more kind of thinking.
> 
> There should also be an `.off` method or the `.on` method should return an "unsubscribe" function. This is important to avoid "leaks" and allow rerender without duplicating the listener.

Great input! thanks.

I've add an option to unsubscribe. And removed the `hooks` parameter. We can add it back in if it becomes necessary.

pavkam (commented):
Should we allow for “before” hooks? Is there any use case where someone might want to add custom data with each track event for example? Just thinking out loud. Also, how do we handle the cases when an error happens. We just don’t trigger the hook now. Should we have an “error”? In case people want to hook it to some custom lib?

roncohen:
> Should we allow for “before” hooks? Is there any use case where someone might want to add custom data with each track event for example? Just thinking out loud. Also, how do we handle the cases when an error happens. We just don’t trigger the hook now. Should we have an “error”? In case people want to hook it to some custom lib?

I think these are valid ideas. Let's take this as a first step. It's basically just the foundation to build some integrations. When the next use-cases show up, we can add more infrastructure  👍

pavkam:
> > Should we allow for “before” hooks? Is there any use case where someone might want to add custom data with each track event for example? Just thinking out loud. Also, how do we handle the cases when an error happens. We just don’t trigger the hook now. Should we have an “error”? In case people want to hook it to some custom lib?
> 
> 
> 
> I think these are valid ideas. Let's take this as a first step. It's basically just the foundation to build some integrations. When the next use-cases show up, we can add more infrastructure  👍 

As long as we don't have to rename in the future it looks great so far

pavkam (approved):

roncohen:
I think `configCheck` and `isEnabledCheck` is better than `checkConfig` and `checkIsEnabled`. Gonna go with that unless someone objects!

Swiftwork (approved):

PR #329, opened 2025-02-20
feat(react-sdk): add onFeaturesUpdated prop and useClient hook

pavkam:
- Introduce `onFeaturesUpdated` callback in BucketProvider to allow custom handling of feature updates
- Add new `useClient()` hook to access the BucketClient directly
- Refactor hooks to use `useClient()` for simplified client access
- Update README with documentation for new features

### DIFF
--- a/packages/react-sdk/README.md
+++ b/packages/react-sdk/README.md
@@ -145,7 +145,6 @@ Example with all options:
 - `appBaseUrl`: Optional base URL for the Bucket application. Use this to override the default app URL,
 - `sseBaseUrl`: Optional base URL for Server-Sent Events. Use this to override the default SSE endpoint,
 - `debug`: Set to `true` to enable debug logging to the console,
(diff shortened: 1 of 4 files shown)


roncohen:
`useClient` is great! And I really like the refactor to remove `ensureProvider` - lets get that in!

I don't particularly like exposing  `onFeaturesUpdated` directly. It's very specific. Since we have `useClient()` people can achieve the same with a `on` with what they get from `useClient()` and have access to all the events? I say let's skip that for now, but get the rest in!

pavkam:
> `useClient` is great! And I really like the refactor to remove `ensureProvider` - lets get that in!
> 
> I don't particularly like exposing `onFeaturesUpdated` directly. It's very specific. Since we have `useClient()` people can achieve the same with a `on` with what they get from `useClient()` and have access to all the events? I say let's skip that for now, but get the rest in!

removed the callback. Give it a 👍🏽 when you can

PR #333, opened 2025-02-21
Small toolbar polish

laander:
https://github.com/user-attachments/assets/9d45f398-933e-4b05-9e7f-5bda3f5599f8

https://github.com/user-attachments/assets/b8e1b30b-5384-42bc-9ad7-7a380d48375b

- added stagger animation for feature rows
- toolbar button and popover is now semi-transparent background with blurred glass effect
- tweaked borders to be inset shadows instead
- made toolbar button round
- fixed single pixel resizing when showing Reset

### DIFF
--- a/packages/browser-sdk/index.html
+++ b/packages/browser-sdk/index.html
@@ -7,7 +7,7 @@
     <meta name="color-scheme" content="light dark" />
     <title>Bucket Browser SDK</title>
   </head>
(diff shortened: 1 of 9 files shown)


roncohen (approved):
nice! we probably need a way to handle it if there are hundreds of features, but we can do that in a separate PR.

PR #342, opened 2025-02-26
docs(browser-sdk): make feature overrides internal for now

roncohen:
We'll be moving exposing feature overrides through what is returned from the `getFeature()` call rather than exposing it directly on the client.

### DIFF
--- a/packages/browser-sdk/src/feature/features.ts
+++ b/packages/browser-sdk/src/feature/features.ts
@@ -20,6 +20,7 @@ export type FetchedFeature = {
 
   /**
    * Result of feature flag evaluation.
+   * Note: does not take local overrides into account.
    */
   isEnabled: boolean;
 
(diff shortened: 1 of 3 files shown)


packages/browser-sdk/src/feature/features.ts
```diff
@@ -71,9 +72,11 @@ export type FetchedFeature = {
 
 const FEATURES_UPDATED_EVENT = "featuresUpdated";
 
+/**
+ * @internal
+ */
 export type FetchedFeatures = Record<string, FetchedFeature | undefined>;
 
-// todo: on next major, come up with a better name for this type. Maybe `LocalFeature`.
 export type RawFeature = FetchedFeature & {
```
roncohen, line 80: I couldn't come up with something better here. I also think we renamed the other type to "FetchedFeature" since this comment was written, so hopefully this is OK now.

Swiftwork (approved):

PR #345, opened 2025-02-26
New Bucket CLI

Swiftwork:
## CLI Package Overview
The [@bucketco/cli](https://www.npmjs.com/package/@bucketco/cli) provides a command-line interface for interacting with the Bucket service. It enables feature flag management and configuration from the terminal.

### Key Features
- **Authentication:** [auth](packages/cli/commands/auth.ts) command for logging in/out of Bucket service
- **Project Setup:** [init](packages/cli/commands/init.ts) command for initializing the Bucket CLI
- **Feature Management:** [features](packages/cli/commands/features.ts) command for listing and creating feature flags
- **App Management:** [apps](packages/cli/commands/apps.ts) command for listing available apps
- **Type Generation:** Automatic TypeScript type generation for feature flags

### Technical Details
- Built with TypeScript and Commander.js for CLI framework
- Uses [ora](https://www.npmjs.com/package/ora) for spinner animations and [chalk](https://www.npmjs.com/package/chalk) for terminal styling
- Configuration stored in `bucket.config.json`
- Auth tokens stored in `~/.bucket-auth` file
- Requires Node.js >= 18.0.0

### Key Commands
```bash
bucket init        # Initialize new configuration
bucket auth login  # Authenticate with Bucket service 
bucket features    # List and manage feature flags
bucket apps list   # List available apps
bucket new         # Create new feature flag
```

The CLI integrates with Bucket's API to provide a streamlined workflow for managing feature flags and configurations from the command line.


packages/cli/utils/gen.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const genDTS = (keys: string[]) => {
+  return /* ts */ `
+// DO NOT EDIT THIS FILE. IT IS GENERATED BY THE BUCKET CLI AND WILL BE OVERWRITTEN.
+// eslint-disable
+// prettier-ignore
+declare module "@bucketco/react-sdk" {
+  interface Features {
+${keys.map((key) => `    "${key}": boolean;`).join("\n")}
```
roncohen, line 77: I think this'll break if you have keys with spaces. It would also be great if we could ensure that this is well formatted so that prettier/linting wont complain. I spent some time on that here: https://github.com/bucketco/bucket-javascript-sdk/blob/feat/cli-minimal/packages/cli/utils/gen.ts#L26-L55

and wrote some tests here: https://github.com/bucketco/bucket-javascript-sdk/blob/feat/cli-minimal/packages/cli/utils/gen.test.ts

roncohen (approved):
this is a great start! Let's get this in and dogfood it to see what else comes up 👍

packages/cli/utils/gen.ts
Swiftwork: 1. I think this'll break if you have keys with spaces.
    - The quotes around keys should allow spaces.
2. It would also be great if we could ensure that this is well formatted so that prettier/linting wont complain.
    - I agree but every customer has different lint and prettier rules so this is an impossibility. We should attempt to format it nicely and then disable the linters like I have done.
3. and wrote some tests here:
    - Planning on writing tests in the next PR as I need to scaffold the whole testing suite which is a lot of files.

Swiftwork:
I'll get this in then to continue writing tests and packaging the solution

packages/cli/utils/gen.ts
Swiftwork: https://github.com/bucketco/bucket-javascript-sdk/pull/345#issuecomment-2693924558

PR #349, opened 2025-02-27
Toolbar: handle many features, tweak stagger anim

laander:
Handle many features:

https://github.com/user-attachments/assets/0f828aca-9544-497a-acdb-02fff4761668

We show a staggered entrance animation for each feature. But if the stagger delay is static and you have a lot of features, that might result in a (too) long animation. So instead, we adjust the stagger delay according to the amount of features in the list so the total animation duration never exceeds 75ms. Here's a demonstration:

https://github.com/user-attachments/assets/ea3c110c-d8e6-443f-b132-8b174d734734

### DIFF
--- a/packages/browser-sdk/src/toolbar/Features.css
+++ b/packages/browser-sdk/src/toolbar/Features.css
@@ -48,15 +48,19 @@
   opacity: 0;
   transform: translateY(-7px);
   transition-property: opacity, transform;
-  transition-duration: calc(0.1s * var(--transition-speed));
+  transition-duration: 0.075s;
   transition-timing-function: cubic-bezier(0.75, -0.015, 0.565, 1.055);
 
   &.show {
     opacity: 1;
     transform: translateY(0);
-    transition-delay: calc(
-      0.075s * var(--transition-speed) / var(--n) * var(--i)
-    );
+/* stagger effect where first item (i=0) has no delay,
+       and delay is based on item count (n) so total animation time always is 509ms */
+    transition-delay: calc(0.05s * var(--i) / max(var(--n) - 1, 1));
+}
+
+  &.hide {
+    visibility: hidden;
   }
 }
 
--- a/packages/browser-sdk/src/toolbar/Features.css
+++ b/packages/browser-sdk/src/toolbar/Features.css
@@ -44,7 +44,6 @@
 }
 
 .feature-row {
-  --transition-speed: 1;
   opacity: 0;
   transform: translateY(-7px);
   transition-property: opacity, transform;
@@ -54,13 +53,9 @@
   &.show {
     opacity: 1;
     transform: translateY(0);
-/* stagger effect where first item (i=0) has no delay,
+    /* stagger effect where first item (i=0) has no delay,
        and delay is based on item count (n) so total animation time always is 509ms */
     transition-delay: calc(0.05s * var(--i) / max(var(--n) - 1, 1));
-}
-
-  &.hide {
-    visibility: hidden;
   }
 }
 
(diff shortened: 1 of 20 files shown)


roncohen:
how about limiting the height/number of shown features to some static number? A huge list isn't useful - you'll use the search anyway. Also, the huge lists takes up a ton of real estate and messes with the feeing that the toolbar is small and light weight. Keeping it limited in height helps it to feel light weight even when you have a ton of features.

packages/browser-sdk/src/toolbar/Toolbar.tsx
```diff
@@ -76,6 +76,10 @@ export default function Toolbar({
   const searchedFeatures =
     search === null ? features : features.filter((f) => f.key.includes(search));
 
+  const sortedFeatures = searchedFeatures.sort((a, b) =>
+    a.key.localeCompare(b.key),
+  );
```
roncohen, line 81: cool but `.sort()` unfortunately updates `searchedFeatures` so the extra variable/assignment just adds noise.

laander: right, went a tad too fast 🙈 fixed

packages/browser-sdk/src/toolbar/Features.css
```diff
@@ -44,16 +44,19 @@
...
+  transition-duration: calc(0.1s * var(--transition-speed));
   transition-timing-function: cubic-bezier(0.75, -0.015, 0.565, 1.055);
 
   &.show {
     opacity: 1;
     transform: translateY(0);
-    transition-delay: calc(0.01s * var(--i));
+    transition-delay: calc(
+      0.075s * var(--transition-speed) / var(--n) * var(--i)
+    );
```
roncohen, line 59: If this was supposed to be max. 0.075s its not doing the intended: 0.075*1/10/9 = 0.00083s, even with 0.075*1/10/10 it only gets to 0.075*1/10/10 = 0.00075

I'd do this instead:`0.075s * var(--i) / var(--n)` which means wait at the most`0.075s` (if --i started from one instead of zero).
And it'd rather go for about half of that: 0.04 to get a snappy feel:

Heres `0.04s` followed by `0.075s`:

https://github.com/user-attachments/assets/5e25d44a-280a-4bd6-8ea2-bc351a28bfe9

I'd also consider decreasing the transition-duration to half:

https://github.com/user-attachments/assets/099c6626-072b-496a-951d-b18a49a03319

laander: > If this was supposed to be max. 0.075s its not doing the intended: 0.0751/10/9 = 0.00083s, even with 0.0751/10/10 it only gets to 0.075*1/10/10 = 0.00075

The last operator is not division ` 0.075*1/10/10` but multiplication ` 0.075*1/10*10`, so it ends up being `= 0,0675`. You're correct, that's 7.5ms off from `0,0750` since it's off by 1 (´--i +1`). but practically that hardly makes a difference .

FWIW `0.075s * var(--i) / var(--n)` is the same as `0.075s / var(--n) * var(--i)` :)

I think your timing tweaks are too aggressive as the stagger effect dissapears almost entirely, but I'll tweak them slightly for a middle-ground.

laander: > I'd do this instead:0.075s * var(--i) / var(--n) which means wait at the most0.075s (if --i started from one instead of zero).

We shouldn't start from `--i + 1` tho as that would result in a delay for the very first item. We want the first item to start animating immediately without delay to not delay the whole animation sequence unnecessarily.
We can however subtract 1 from `--n` instead and cap it at min 1 to avoid division by zero.

Starting to become a lil convoluted, but alas :)

roncohen: > The last operator is not division ` 0.075*1/10/10` but multiplication

doh! I was tired and fighting a cold and had the nagging feeling that I could hurt myself if I tried doing math in that condition. Turns out I was right - and wrong. Apologies for that.

roncohen (approved):

PR #355, opened 2025-03-03
Publish package CLI

Swiftwork:
Updated readme and packaging

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -2,13 +2,16 @@
 
 Command-line interface for interacting with Bucket services. The CLI allows you to manage apps, features, authentication, and generate TypeScript types for your Bucket features. With this tool, you can streamline your feature flagging workflow directly from your terminal.
 
-## Installation
+## Quick Start
 
-You can install the Bucket CLI either locally to your project or globally on your system.
+Get started quickly by running the CLI directly: initializing the CLI, creating a feature, and generate the types all at once.
 
-### Local Installation (Recommended)
+```bash
+# Initialize CLI (if not setup), create a feature, and generate types all at once
+npx @bucketco/cli new
+```
 
-Local installation is recommended for project-specific usage and ensures consistent CLI versions across your team.
+or install it locally
 
 ```bash
 # npm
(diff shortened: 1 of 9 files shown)


packages/cli/package.json
```diff
@@ -1,10 +1,17 @@
 {
   "name": "@bucketco/cli",
-  "version": "0.0.1",
+  "version": "0.1.0",
   "packageManager": "yarn@4.1.1",
   "description": "CLI for Bucket service",
   "main": "./dist/index.js",
   "type": "module",
+  "license": "MIT",
+  "author": "Bucket Co.",
```
roncohen, line 9:
```suggestion
  "author": "Bucket.",
```

packages/cli/README.md
```diff
@@ -1 +1,246 @@
...
+
+# pnpm
+pnpm add --save-dev @bucketco/cli
+```
+
+When installed locally, use the CLI with npx:
+
+```bash
+# npm
+npx bucket <command>
```
roncohen, line 28: You don't even need to install it if you're using `npx` i believe?

packages/cli/README.md
```diff
@@ -1 +1,246 @@
...
+```
+
+or as a `package.json` script
+
+```json
+{
+  "scripts": {
+    "bucket": "bucket"
```
roncohen, line 36: this seems awkward... If you're using `yarn` at least you can just do `yarn bucket` without this, but it does look like `npm` doesn't give you that. Hmmm

packages/cli/README.md
```diff
@@ -1 +1,246 @@
...
+
+# pnpm
+pnpm add -g @bucketco/cli
+```
+
+When installed globally, use the CLI directly:
+
+```bash
+bucket <command>
+```
```
roncohen, line 68: feels like we should take this out. Personally my experience with global installations is that it's more pain than it's worth?

roncohen (approved):
Let's get this published!

packages/cli/package.json
Swiftwork: With the dot?

packages/cli/README.md
Swiftwork: You would have to run it with the full package name without installing it. Would you like me to add that as a suggestion?
```bash
npx @bucketco/cli <command>
```

packages/cli/README.md
```diff
@@ -1 +1,246 @@
...
+
+## Quick Start
+
+Get started quickly with the following commands:
+
+```bash
+# Initialize CLI (if not setup), create a feature, and generate types all at once
+bucket new "My Feature"
```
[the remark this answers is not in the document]
Swiftwork: I can adjust :)

packages/cli/package.json
Swiftwork: Kept the dot

packages/cli/README.md
Swiftwork: Clarified this, need to test to see what is possible when published

packages/cli/README.md
Swiftwork: Removed this and wrote the alternatives.

packages/cli/README.md
Swiftwork: Demphasised and made it an alternative further down.

packages/cli/package.json
roncohen: hah, sorry, we should remove the dot. Apologies!

packages/cli/README.md
roncohen: OK, lets test once the package exists.

packages/cli/README.md
roncohen: perfect, thanks!

PR #360, opened 2025-03-05
fix:(browser-sdk,react-sdk): consolidated specific enabled/config check listeners

roncohen:
- Deprecate enabledCheck and configCheck event listeners in favour of just `check`.
- Publishes new patch version for browser and react sdks.

### DIFF
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -799,6 +799,7 @@ export class BucketClient {
         checkEvent.action == "check-config" ? "configCheck" : "enabledCheck",
         checkEvent,
       );
+      this.hooks.trigger("check", checkEvent);
     });
   }
 
--- a/packages/browser-sdk/src/client.ts
+++ b/packages/browser-sdk/src/client.ts
@@ -480,30 +480,31 @@ export class BucketClient {
   }
 
   /**
-   * Add a hook to the client.
+   * Add an event listener
    *
-   * @param hook Hook to add.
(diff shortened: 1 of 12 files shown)


packages/browser-sdk/src/client.ts
```diff
@@ -480,30 +480,31 @@ export class BucketClient {
...
+   * @param handler The function to call when the event is triggered.
    * @returns A function to remove the hook.
    */
   on<THookType extends keyof HookArgs>(
     type: THookType,
     handler: (args0: HookArgs[THookType]) => void,
   ) {
-    this.hooks.addHook(type, handler);
-    return () => this.hooks.removeHook(type, handler);
+    return this.hooks.addHook(type, handler);
```
roncohen, line 493: `addHook` already returns a function to remove this, in case you're wondering

laander (approved):
👍 

we should document the `CheckEvent` properties somewhere accessible as you need to know that `action` can be used as a discriminator:
```
CheckEvent = {
      action: "check-is-enabled",
      key: "test-key",
      value: true,
    };
```
does this automatically get surfaced on docs.bucket.co when docs are auto-built? we could also just JSDoc clause with a short description

roncohen:
> we should document the `CheckEvent` properties somewhere accessible as you need to know that `action` can be used as a discriminator

agreed. And yeah, we do automatic reference docs now, but those can also be improved. I'll get this in now and look at improving the docs separately.

PR #361, opened 2025-03-06
chore: fail on docs warnings and export missing types

matus-vacula:


roncohen:
Never mind, that doesn't do what I hoped:

Screenshot 2025-03-06 at 12 01 17

roncohen:
actually this might still be better. The idea is to keep what the SDK exposes minimal. We still want it to be documented though, so maybe this is a good middle way.. hmm

roncohen (approved):
never mind, lets just get this in for now.

PR #374, opened 2025-03-13
feat: improve logging by fetching bootstrap first

Swiftwork:
By adding a call to bootstrap we can improve the logging of the CLI.

https://github.com/user-attachments/assets/9ccac815-84da-4591-b4e1-cca86ccb5944

### DIFF
--- a/packages/cli/commands/features.ts
+++ b/packages/cli/commands/features.ts
@@ -9,7 +9,12 @@ import { App, getApp, getOrg } from "../services/bootstrap.js";
 import { createFeature, Feature, listFeatures } from "../services/features.js";
 import { configStore } from "../stores/config.js";
 import { handleError, MissingAppIdError } from "../utils/errors.js";
-import { genFeatureKey, genTypes, KeyFormatPatterns } from "../utils/gen.js";
+import {
+  genFeatureKey,
+  genTypes,
+  indentLines,
+  KeyFormatPatterns,
+} from "../utils/gen.js";
 import {
   appIdOption,
   featureKeyOption,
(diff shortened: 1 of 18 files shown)


packages/cli/commands/features.ts
```diff
@@ -46,13 +52,11 @@ export const createFeatureAction = async (
...
 
-    spinner = ora(
-      `Creating feature for app ${chalk.cyan(appId)} at ${chalk.cyan(baseUrl)}...`,
-    ).start();
+    spinner = ora(`Creating feature...`).start();
     const feature = await createFeature(appId, name, key);
     // todo: would like to link to feature here but we don't have the env id, only app id
     spinner.succeed(
-      `Created feature ${chalk.cyan(feature.name)} with key ${chalk.cyan(feature.key)} at ${chalk.cyan(baseUrl)}.`,
+      `Created feature ${chalk.cyan(feature.name)} with key ${chalk.cyan(feature.key)}${baseUrlSuffix(baseUrl)}.`,
```
roncohen, line 59: can we link to the feature on bucket?

roncohen (approved):
Generally i think it's great to have improved logging 👍  A little worried about the bootstrap call what we'll incur on every invocation in terms of just making the CLI feel snappy, but it's probably fine.

One general comment: I don't think we need to log out request by default. In the video it logs "Loaded apps from ...". Thats probably unnecessary. 

Btw. if you always pull the apps, we don't need to store the featureKeyFormat anymore since that's available in the bootstrap. Interestingly, it's at the Org level. I would have expected it to be at the app level, but it'll have to do for now.

packages/cli/commands/features.ts
Swiftwork: Not without envId or going back to appId in the URL (which I think is the plan long term)

Swiftwork:
Based on your feedback I rewrote the code in https://github.com/bucketco/bucket-javascript-sdk/pull/374/commits/52e77da65bfa5c59f143fb37aebd60a40820a69a to load feature key and to write out the entire URL to the new feature as we now have the environments loaded.
image

packages/cli/commands/features.ts
Swiftwork: Fixed

PR #376, opened 2025-03-14
docs(cli,react-sdk,node-sdk): Introduce the CLI for type gen

roncohen:
This updates the default workflow for React to use the CLI to remove friction and generate types. 
The Node.js docs still need some more love, but this is a start.


fmerian (approved):
lgtm!

PR #384, opened 2025-03-23
feat: use webcrypto API to support edge runtimes

roncohen:
Since `digest` is async, we need to update the interface to be async.

### DIFF
--- a/packages/flag-evaluation/src/index.ts
+++ b/packages/flag-evaluation/src/index.ts
@@ -1,4 +1,10 @@
-import { createHash } from "node:crypto";
+try {
+  // crypto not available on globalThis in Node.js v18
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  globalThis.crypto ??= require("node:crypto").webcrypto;
+} catch {
+  // ignore
+}
 
 /**
  * Represents a filter class with a specific type property.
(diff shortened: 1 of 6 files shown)


packages/flag-evaluation/src/index.ts
```diff
@@ -366,30 +376,38 @@ function evaluateRecursively(
...
-          acc || evaluateRecursively(current, context, missingContextFieldsSet)
+
+        const newRes = await evaluateRecursively(
+          current,
+          context,
+          missingContextFieldsSet,
         );
-      }, filter.operator === "and");
+
+        result = isAnd ? result && newRes : result || newRes;
```
roncohen, line 401: @matus-vacula rewrote to simple iterative approach because it was simpler with the switch to async/await. Could you double check my logic here?

roncohen:
since this changes flag evalution to async, the Node SDK need to be updated separately to also use async when calling `getFeatures` which requires a new major version :/

packages/flag-evaluation/src/index.ts
matus-vacula: This seems correct.

matus-vacula:
Is the purpose of this to support the local evaluation in react and browser SDK? If we want to avoid the need for breaking the backwards compatibility for NodeSDK we can have 2 different implementations (sync and async) theoretically.

PR #385, opened 2025-03-23
feat(node-sdk): add `getFeatureDefinitions`

roncohen:
Introduces a new type `FeatureDefinition` which describes the feature, including rules for isEnabled and config as well as the description.

### DIFF
--- a/packages/node-sdk/src/types.ts
+++ b/packages/node-sdk/src/types.ts
@@ -191,6 +191,56 @@ export type FeatureOverride =
   | boolean;
 
 /**
+ * Describes a feature definition.
+ */
+export type FeatureDefinition = {
+  /**
+   * The key of the feature.
+   */
+  key: string;
+
+  /**
+   * Description of the feature.
+   */
+  description: string | null;
+
+  /**
+   * The targeting rules for the feature.
+   */
+  isEnabled: {
+    /**
+     * The version of the targeting rules.
+     */
+    version: number;
+
+    /**
+     * The targeting rules.
+     */
+    rules: {
+      /**
+       * The filter for the rule.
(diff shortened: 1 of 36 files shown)


packages/node-sdk/src/types.ts
```diff
@@ -190,6 +190,56 @@ export type FeatureOverride =
...
+
+  /**
+   * Description of the feature.
+   */
+  description: string | null;
+
+  /**
+   * The targeting rules for the feature.
+   */
+  isEnabled: {
```
roncohen, line 210: this is `targeting` in the API, but that didn't seem right so I've exposed this as `isEnabled`. Suggestions are welcome!

roncohen:
> > Introduces a new type `FeatureDefinition` which describes the feature, including rules for isEnabled and config as well as the description.
> 
> What's the envisioned use case for this?

We're making it possible to retrieve a bunch of meta data about features to be able to make this data available in something like Vercel's Flags SDK.

packages/node-sdk/src/types.ts
pavkam: `access`? And have the `isEnabled: true` inside maybe?

roncohen: I don't like access because we're increasingly heard that "access" doesn't really cover turning the thing on/off for companies. Leaning towards `toggleTargeting` and `configTargeting` atm..

pavkam (approved):

packages/node-sdk/src/types.ts
pavkam: I approved and resolving this comment as it's being discussed internally.

PR #388, opened 2025-03-25
fix(node-sdk): increase default `get` timeout to 10s, add 3 retries

roncohen:
Our HTTP API responds to `/feature` requests in less than 5ms, but there are cases where the network on the client could be dodgy or there's otherwise some delay. Especially when the program is just starting up this is problematic because the NodeSDK will use the fallback flags which for the most part are empty.

This introduces a configurable timeout and default the timeout to 10s instead of the existing 5s. It also adds an exponential retry which is set to 3 tries by default.

### DIFF
--- a/packages/node-sdk/package.json
+++ b/packages/node-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bucketco/node-sdk",
-  "version": "1.6.4",
+  "version": "1.6.5",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


pavkam (commented):
Approving but we should use some off the shelf policy management library that offers the retry timeout and back off stuff

roncohen:
> Approving but we should use some off the shelf policy management library that offers the retry timeout and back off stuff

thanks! I took a quick look to try and find an appropriate library. Wanted:
- simple
- popular, ideally maintained by someone serious
- uses `unref` on timers to avoid keeping the server running if everything else is done

I didn't find anything immediately and this fixes a current issue for customers so I'll go ahead and merge, making a mental note to be on the lookout for an appropriate library to replace our own implementation.

pavkam (approved):

PR #412, opened 2025-05-20
feat: add client and client version to request headers

Swiftwork:


pavkam (approved):

Swiftwork:
> removing from queue to ensure we get this discussion in ^^

Good.

> The SDKs are using [`bucket-sdk-version`](https://github.com/bucketco/bucket-javascript-sdk/blob/6747e5a9346177f87dfee081cfdf1b13674da77a/packages/browser-sdk/src/config.ts#L7). Maybe it makes sense for the CLI as well?

Hmm, good point.

Swiftwork:
> The SDKs are using [`bucket-sdk-version`](https://github.com/bucketco/bucket-javascript-sdk/blob/6747e5a9346177f87dfee081cfdf1b13674da77a/packages/browser-sdk/src/config.ts#L7). Maybe it makes sense for the CLI as well?

Out of curiosity, is there a reason we didn't go for an `X-` header? And should I use a similar `bucket-sdk-client` header alongside it @roncohen ?

roncohen:
AFAIR we investigated and found that there was no additional benefit of prefixing "x-". E.g. the pattern was recommended against as outdated. But I admit it's all from hazy memory.

Swiftwork:
> AFAIR we investigated and found that there was no additional benefit of prefixing "x-". E.g. the pattern was recommended against as outdated. But I admit it's all from hazy memory. 

Fair enough. I'll update the code to use `bucket-sdk-version` and introduce `bucket-sdk-client`. I think I must update the web to accept this new header first so not reject requests.

roncohen:
> I'll update the code to use `bucket-sdk-version` and introduce `bucket-sdk-client`

We've sending both in the same header following this pattern today: `client/version`. See  https://github.com/bucketco/bucket-javascript-sdk/blob/main/packages/browser-sdk/src/config.ts#L9 for an example.

Swiftwork:
> > I'll update the code to use `bucket-sdk-version` and introduce `bucket-sdk-client`
> 
> We've sending both in the same header following this pattern today: `client/version`. See https://github.com/bucketco/bucket-javascript-sdk/blob/main/packages/browser-sdk/src/config.ts#L9 for an example.

Done :)

roncohen:
thanks @Swiftwork. I just realized the CLI is of course talking to the REST API, not the front facing API, so there's no actual requirement to follow what the SDKs do - apologies for the turkey chase. I'm OK to go with `bucket-sdk-client` now since it'll at least be consistent with the front facing API.

roncohen (approved):

packages/cli/stores/config.ts
```diff
@@ -101,6 +102,20 @@ class ConfigStore {
       const content = await readFile(this.configPath, "utf-8");
       const parsed = parseJSON(content) as unknown as Partial<Config>;
 
+      if (packageJSONPath) {
```
roncohen, line 105: will this find the CLI package json or the project package JSON?

roncohen: can we test this out end to end to make sure?

Swiftwork: Right, good question. I think the later... When testing it worked ofc as `package.json` is the same but bundled it won't be, but I'll take another look. Annoying we can't directly import json files without the experimentation warning like we do in the other SDKs.

Swiftwork: I have resolved this and tested end-to-end locally with https://github.com/bucketco/bucket-web/pull/2634

packages/cli/stores/config.ts
```diff
@@ -89,11 +91,25 @@ class ConfigStore {
...
     try {
-      const packageJSONPath = await findUp("package.json");
+      const moduleMetadata = await readFile(
+        join(moduleRoot, "package.json"),
+        "utf-8",
+      );
+      const moduleMetadataParsed = parseJSON(moduleMetadata) as unknown as {
+        version: string;
+      };
+      this.clientVersion = moduleMetadataParsed.version;
```
roncohen, line 103: this is probably fine, but as a point of reference this is what the SDKs do: https://github.com/bucketco/bucket-javascript-sdk/blob/main/packages/browser-sdk/src/config.ts#L1

Swiftwork: We can't unfortunately do that without getting the "importing json is experimental warning" because the package uses type "module"

PR #424, opened 2025-06-18
Fix numeric comparison in evaluate bug

laander:
In local evaluation for the Node SDK:

Fix numeric comparison for GT/LT operators where casting strings to numbers wasn't being done correctly happen.
That could result in situations where `"2" > "10" == true`

Added test case

### DIFF
--- a/packages/flag-evaluation/src/index.ts
+++ b/packages/flag-evaluation/src/index.ts
@@ -304,7 +304,7 @@ export function evaluate(
         );
         return false;
       }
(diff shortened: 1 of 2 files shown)


roncohen (approved):

laander:
Superceded by https://github.com/bucketco/bucket-javascript-sdk/pull/426 because this PR was created from another remote (forked repo) which meant that GH actions build and test wouldn't run

PR #453, opened 2025-08-04
feat: add new context filter operators

Swiftwork:
- Bumped version of `@bucketco/flag-evaluation` to 0.2.0 and added new context filter operators: `DATE_AFTER` and `DATE_BEFORE`.
- Updated `@bucketco/node-sdk` to 1.9.0 to reflect the new version of `@bucketco/flag-evaluation`.
- Updated `@bucketco/openfeature-node-provider` to use the latest `@bucketco/node-sdk` version.

### DIFF
--- a/packages/flag-evaluation/src/index.ts
+++ b/packages/flag-evaluation/src/index.ts
@@ -344,15 +344,19 @@ export function evaluate(
         ? fieldValueDate > daysAgo.getTime()
         : fieldValueDate < daysAgo.getTime();
     }
-    case "DATE_AFTER": {
-      const fieldValueDate = new Date(fieldValue).getTime();
-      const valueDate = new Date(value).getTime();
-      return fieldValueDate >= valueDate;
-    }
+    case "DATE_AFTER":
     case "DATE_BEFORE": {
       const fieldValueDate = new Date(fieldValue).getTime();
       const valueDate = new Date(value).getTime();
-      return fieldValueDate <= valueDate;
+      if (isNaN(fieldValueDate) || isNaN(valueDate)) {
+        console.error(
+          `${operator} operator requires valid date values: ${fieldValue}, ${value}`,
+        );
+        return false;
+      }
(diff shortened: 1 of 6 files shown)


packages/flag-evaluation/src/index.ts
```diff
@@ -342,6 +344,16 @@ export function evaluate(
         ? fieldValueDate > daysAgo.getTime()
         : fieldValueDate < daysAgo.getTime();
     }
+    case "DATE_AFTER": {
```
pavkam, line 347: you can probably combine these two like in the case of `AFTER` and `BEFORE`. And only change the operator

packages/flag-evaluation/src/index.ts
```diff
@@ -342,6 +344,16 @@ export function evaluate(
...
+      const fieldValueDate = new Date(fieldValue).getTime();
+      const valueDate = new Date(value).getTime();
+      return fieldValueDate >= valueDate;
+    }
+    case "DATE_BEFORE": {
```
pavkam, line 352: Needs tests

packages/flag-evaluation/src/index.ts
Swiftwork: Good point

packages/flag-evaluation/src/index.ts
Swiftwork: Will do

packages/flag-evaluation/src/index.ts
```diff
@@ -342,6 +344,16 @@ export function evaluate(
         ? fieldValueDate > daysAgo.getTime()
         : fieldValueDate < daysAgo.getTime();
     }
+    case "DATE_AFTER": {
+      const fieldValueDate = new Date(fieldValue).getTime();
+      const valueDate = new Date(value).getTime();
```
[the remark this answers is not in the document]
Swiftwork: Done

PR #455, opened 2025-08-06
fix(node-sdk): avoid sending "check"'s when using getFeatures

roncohen:
The `getFeatures` endpoint is mostly used to serialize flags for the frontend. When that happens, "checks" will be sent for all features.

### DIFF
--- a/packages/openfeature-node-provider/package.json
+++ b/packages/openfeature-node-provider/package.json
@@ -50,7 +50,7 @@
     "vitest": "~1.6.0"
   },
   "dependencies": {
-    "@bucketco/node-sdk": "1.9.2"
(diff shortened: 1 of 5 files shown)


packages/openfeature-node-provider/package.json
```diff
@@ -50,7 +50,7 @@
     "vitest": "~1.6.0"
   },
   "dependencies": {
-    "@bucketco/node-sdk": "1.9.2"
+    "@bucketco/node-sdk": "1.9.3"
```
pavkam, line 53: You need to up the package version as well

pavkam (approved):

PR #458, opened 2025-08-13
fix: make stop propagated event listeners passive

Swiftwork:


Swiftwork:
> Looks fine, but I'll trust you know what you are doing :)

Did some tests and passive does not impact `stopPropgation`. Which is the only purpose of these code paths.

PR #478, opened 2025-09-17
fix: version check can cause failures

roncohen:
If the version check is blocked for some reason, the process can fail. Apparently promise rejections can happen without `await` calls.

### DIFF
--- a/packages/cli/utils/version.ts
+++ b/packages/cli/utils/version.ts
@@ -25,41 +25,36 @@ export async function current() {
 }
 
 async function getLatestVersionFromNpm(packageName: string): Promise<string> {
-  // await new Promise((resolve) => setTimeout(resolve, 1000));
-  throw new Error(`Failed to fetch latest version from npm: ${packageName}`);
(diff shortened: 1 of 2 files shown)


packages/cli/utils/version.ts
```diff
@@ -25,37 +25,41 @@ export async function current() {
 }
 
 async function getLatestVersionFromNpm(packageName: string): Promise<string> {
-  try {
-    const response = await fetch(`https://registry.npmjs.org/${packageName}`, {
-      signal: AbortSignal.timeout(5000),
-    });
+  // await new Promise((resolve) => setTimeout(resolve, 1000));
```
pavkam, line 28: Not sure why this happens. If we catch failed promises, even if now awaited, should not fail. What exactly is happening?

I'm not against removing the try but need to understand why rejections aren't processed.

roncohen: we did catch it, but then threw again inside the catch. That second one wasn't being catch'ed.

